PR #9, opened 2015-04-02
Suppress MSVC warning

MarcelRaad:
warning C4244: 'argument': conversion from 'int' to 'char', possible loss of data

This causes the Boost.Units regression tests to fail on MSVC because of warnings-as-errors:
http://www.boost.org/development/tests/develop/developer/output/teeks99-08f-win2012R2-64on64-boost-bin-v2-libs-units-test-test_information_units-test-msvc-14-0-dbg-adrs-mdl-64-archt-x86-async-excpt-on-thrd-mlt.html


jzmaddock:
Thanks!

PR #11, opened 2015-07-23
Fix Miller-Rabin not recognizing 2 as prime.

TPolzer:
miller_rabin_test returned false for all even numbers, including 2.


jzmaddock:
I fixed this by just adding a special case for n == 2: https://github.com/boostorg/multiprecision/commit/9e3115f376d01ff0e0a51d9ad0a487533c0689c2

PR #15, opened 2016-06-07
Fix for MSVC SmallerTypeCheck runtime error

fun4jimmy:
The assignment of carry to p in eval_multiply generates a runtime error in non-optimized MSVC builds which have SmallerTypeCheck enabled. As carry is twice the size of p and the top bits are non-zero the check errors saying that some bits are being ignored. This is Microsoft's recommended fix of masking the unwanted bits when assigning to the smaller type.

### DIFF
--- a/include/boost/multiprecision/cpp_int/multiply.hpp
+++ b/include/boost/multiprecision/cpp_int/multiply.hpp
@@ -36,7 +36,7 @@ inline typename enable_if_c<!is_trivial_cpp_int<cpp_int_backend<MinBits1, MaxBit
    while(p != pe)
    {
       carry += static_cast<double_limb_type>(*pa) * static_cast<double_limb_type>(val);
-      *p = static_cast<limb_type>(carry);
+      *p = carry & ~static_cast<limb_type>(0);
       carry >>= cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>::limb_bits;
       ++p, ++pa;
    }


fun4jimmy:
There is a` #pragma runtime_checks` which I've tried to use as a temporary workaround but I can't seem to get it working with a generic template function. I've found [this stackoverflow question](http://stackoverflow.com/questions/28473206/how-to-disable-runtime-checks-for-function-templates-in-visual-studio-c) that seems to be asking about the same problem but there are no answers as to how to work around it.

fun4jimmy:
The code generating the runtime error is:
```cpp
const int64_t Input = 135000000;
const uint32_t Multiplier = 135000000;
boost::multiprecision::int128_t Value = Input;
Value *= Multiplier;
```

jzmaddock:
This is the tip of a very large iceberg - running the regression tests with /RTCc results in hundreds of these issues - if I can fix them I will, but I'd prefer not to add unnecessary bitmask operations just to keep MSVC happy.  I realize msvc will optimize them away, whether all compilers do so is another matter...

fun4jimmy:
Ok, I understand. Let me know what you decide to do either way.

I'll open a ticket with Microsoft to get them to fix disabling the warning for template functions using `#pragma runtime_checks` so it can be worked around if necessary. I'll add a link to the issue here once it's opened.

Thanks,
James

jzmaddock:
Fixed in https://github.com/boostorg/multiprecision/commit/c7e29b02ff696f01e52c0933ff3cf5c3b9bd88fe

PR #17, opened 2016-11-08
[inspection report] Remove tabs and prevent min macro expansion

zerotypos-found:


zerotypos-found:
Abandoned. Please See #18.
Thanks.

PR #18, opened 2016-11-14
[inspection report] Remove tabs and protect std::min from macro expansion

zerotypos-found:
The previous PR consists of multiple commits with bad commit messages (which was done using GitHub Web-interface editing). Thus, squashed and improved the commit log.


jzmaddock:
Thanks!

Issue #19, opened 2017-03-20
Huge difference between cpp_dec_float_100 and double


andrew-vandenbrink:
```
typedef boost::multiprecision::cpp_dec_float_100 decimal;
decimal IRR = decimal(0.0227);
auto x = decimal(-105.0) + decimal(2.0) * (decimal(1.0) / (IRR / decimal(2.0)) - decimal(1.0) / (IRR / decimal(2.0))*(decimal(1.0) / pow((decimal(1.0) + IRR / decimal(2.0)), decimal(6.0)))) + decimal(100.0) / pow((decimal(1.0) + IRR / decimal(2.0)), decimal(6.0));
//x=111.537
```

I replace decimal with double with exact and same formula
```
double IRR = 0.0227;
auto x = double(-105.0) + double(2.0) * (double(1.0) / (IRR / double(2.0)) - double(1.0) / (IRR / double(2.0))*(double(1.0) / pow((double(1.0) + IRR / double(2.0)), double(6.0)))) + double(100.0) / pow((double(1.0) + IRR / double(2.0)), double(6.0));
//x=-0.0100884
```
I get the same result with double if I use excel.
Is there a bug in my formula when I use decimal or is there any other issue?

ckormanyos:
> Is there a bug in my formula when I use decimal or> is there any other issue?
I am looking into this.
Thanks for the report.
Best regards, Chris
 

    On Monday, March 20, 2017 7:41 PM, andrew-vandenbrink <notifications@github.com> wrote:
 

 typedef boost::multiprecision::cpp_dec_float_100 decimal;
decimal IRR = decimal(0.0227);
auto x = decimal(-105.0) + decimal(2.0) * (decimal(1.0) / (IRR / decimal(2.0)) - decimal(1.0) / (IRR / decimal(2.0))*(decimal(1.0) / pow((decimal(1.0) + IRR / decimal(2.0)), decimal(6.0)))) + decimal(100.0) / pow((decimal(1.0) + IRR / decimal(2.0)), decimal(6.0));
//x=111.537
I replace decimal with double with exact and same formuladouble IRR = 0.0227;
auto x = double(-105.0) + double(2.0) * (double(1.0) / (IRR / double(2.0)) - double(1.0) / (IRR / double(2.0))*(double(1.0) / pow((double(1.0) + IRR / double(2.0)), double(6.0)))) + double(100.0) / pow((double(1.0) + IRR / double(2.0)), double(6.0));
[...]

PR #25, opened 2017-06-01
Store basic arithmetic types by value in expression templates.

jzmaddock:
So that we can't get a dangling-reference to a literal if someone stores an expression, see https://github.com/boostorg/multiprecision/issues/19.


ckormanyos:
Thanks John!

PR #26, opened 2017-06-02
Remove keyword template in non-template context.

mglisse:
Not sure if this is the source or a generated file.


jzmaddock:
Generated file, I'll fix this manually in a moment, closing for now to prevent this generating a needless CI build.

jzmaddock:
Fixed in https://github.com/boostorg/multiprecision/commit/63488e1a0e0bba8d80dbbc4b8642a452b3155f8c

Issue #43, opened 2018-02-04
0!=0


ohhmm:
try this test
    boost::multiprecision::cpp_int i = 0;
    BOOST_TEST(i==0);
    ++i--;
    BOOST_TEST(i==0);
it fails in 1.65 because operator== compares limbs count.

jzmaddock:
Yup, eval_increment/decrement don't set the sign to zero when the result yields zero, testing a fix now.

PR #44, opened 2018-02-05
Add test case for #43 Multiprecision: cpp_int 0!=0

ohhmm:
Multiprecision: cpp_int::operator== compares size() which return limbs count that is zero for just initialized zero value but becomes one after some operations. If some operations returns value to zero, limbs count is not zeroed. This may be fixed in different ways. Adding the test for now.


ohhmm:
Also take a look onto eval_increment fix: check for becoming zero to change sign.

PR #50, opened 2018-03-15
Fix infinite loop in gmp_float to fixed-point string conversion

emfrias:
You can trigger the infinite loop with: 
```cpp
cout << setprecision(1) << fixed << mpf_float_50("0.0501");
```


jzmaddock:
Thanks for that!

Issue #67, opened 2018-08-16
Multiprecision get_constant_one_over_epsilon() not thread safe


mkorpar:
```
template <class T>
const T& get_constant_one_over_epsilon()
{
   static const bool is_init = false;
   static T result;
   if (is_init == false)
   {
      typedef typename mpl::front<typename T::unsigned_types>::type ui_type;
      result = static_cast<ui_type>(1u);
      eval_divide(result, std::numeric_limits<number<T> >::epsilon().backend());
   }

   constant_initializer<T, &get_constant_one_over_epsilon<T> >::do_nothing();

   return result;
}
```

is_init is never set to true.

jzmaddock:
Not sure what happened there, none of that initialization code was quite right.  Should all be consistent and fixed now.

PR #68, opened 2018-08-16
make get_constant_one_over_epsilon thread safe

mkorpar:
reimplement get_constant_one_over_epsilon to use same methods as other constant methods

https://github.com/boostorg/multiprecision/issues/67


jzmaddock:
Thanks for that, I went for a slightly different fix.

PR #86, opened 2018-09-18
Narrowing ref-qualifiers to compile on VS2012/2013

raffienficiaud:
This should fix a compilation issue on VS2012/2013, although I have not tested it.


raffienficiaud:
Tests are shown as failing: 

* on Travis, I think a worker died
* on Appveyor, there is a compilation issue? I did not get it very well.

jzmaddock:
Good catch - that's the right fix, the CI failures are unrelated.

PR #89, opened 2018-09-26
Use allocator_traits for rebind when appropriate via rebind helper

glenfe:


glenfe:
No more single failure. Everything passes after your "Fix up assignment from string_view for mpf_float" commit.

jzmaddock:
Good catch - thanks for this!

PR #90, opened 2018-10-01
Support allocators declared as final

glenfe:
Use `boost::empty_value<A>`  which does the right thing if `A` is `final`.


jzmaddock:
Looks fine - thanks again!

PR #95, opened 2018-11-04
Fix compilation failure with clang C++03 and Boost.Optional

Lastique:
This PR fixes compilation failures with clang in C++03 mode that were found in https://github.com/boostorg/integer/pull/11#issuecomment-435649481. Two problems are fixed here:

- `is_explicitly_convertible` trait required a public destructor in C++03 mode because `declval` returns `T` by value in this mode. The compilation fails because `optional_base`, on which the trait gets instantiated during `optional` constructor compilation, has a protected destructor.
- `is_byte_container` trait required a public `value_type` typedef in the range type. `optional_base` defines a protected typedef `value_type`. In general, this typedef is quite commonly used in non-range types, so don't use it to detect container/range types.

More details are available in commit messages.

Also, the PR adds a few missing includes and fixes a gcc warning about implicit fallthough in switch/case. It also adds a test for compatibility between Boost.Multiprecision and Boost.Optional.

### DIFF
--- a/include/boost/multiprecision/detail/functions/pow.hpp
+++ b/include/boost/multiprecision/detail/functions/pow.hpp
@@ -537,6 +537,7 @@ inline void eval_pow(T& result, const T& x, const T& a)
          {
             // fallthrough..
          }
+         BOOST_FALLTHROUGH;
       }
       default:
          if(eval_signbit(a))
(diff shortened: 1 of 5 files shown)


jzmaddock:
Thanks for sorting that one out!  Merged.

PR #102, opened 2018-12-17
fix warning about unused variable, closes #92

### LINKED ISSUE #92: cpp_int.hpp, unused parameter warning in clang
cpp_int.hpp:833
`BOOST_MP_FORCEINLINE void resize(unsigned new_size, unsigned min_size)`

raises a warning on clang compilers, since new_size is not used inside the function. Please change this to

`BOOST_MP_FORCEINLINE void resize(unsigned /* new_size */, unsigned min_size)`

HDembinski:


HDembinski:
I am using cpp_int in boost.histogram, and currently I am turning of the warning with a pragma in my code. Would be nice to fix this upstream :)

Issue #113, opened 2019-02-12
wrong number of decimal places formatting mpf_float as fixed-point


emfrias:
Hello,

I've encountered a bug where I get the wrong number of digits when formatting an mpf_float_50 as fixed precision:
```
std::cout << std::setprecision(3) << std::fixed << boost::multiprecision::mpf_float_50("99.9809") << "\n";
```
This prints "99.9809" instead of the expected "99.981".

I traced into it a little and it looks like the code starts by [calling `mpf_get_str()`](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/gmp.hpp#L292)  asking for three digits,
which returns `"1"` with `e == 3` (meaning 100).  So it turns around and [calls mpf_get_str()](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/gmp.hpp#L345) again, this time asking
for 6 digits, and this time it returns `"999809"` with `e == 2` (meaning 99.9809) -- the extra digits caused `mpf_get_str()` to round differently.

I've hacked my copy to check whether `e` changes, and if so, adjusts `digits` accordingly and calling `mpf_get_str()` a third time, but I'm not sure that's the right solution.

jzmaddock:
Will be in the next release.

PR #118, opened 2019-02-22
Fix bug in gmp float -> fixed point string formatting

emfrias:
proposed fix for #113


jzmaddock:
Thanks for that, testing enhanced tests plus fix now...

PR #126, opened 2019-03-26
Don't construct stringstream multiple times

zmij:
It's inefficient to construct the stream object for each element in the array


ckormanyos:
Good catch. Good idea. Thank you.

ckormanyos:
Improve efficiency by not creating stringstream object multiple times in printing loop.

PR #130, opened 2019-04-28
Respect uppercase for hexadecimal outputs

### LINKED ISSUE #129: String output does not respect uppercase flag for hexadecimal base prefix
Consider following program:

```c++
#include <cstdint>

int main()
{
    uint64_t val64 = 0xf;
    boost::multiprecision::uint128_t val128 = 0xf;
    std::cout << std::hex << std::uppercase << std::showbase
              << "uint64: "  << val64 << std::endl
              << "uint128: " << val128 << std::endl;
    return 0;
}
```

The output is:
```
uint64: 0XF
uint128: 0xF
```

The expected output is (note the uppercase for `0x`):
```
uint64: 0XF
uint128: 0XF
```

pavelkryukov:
Fixes #129


jzmaddock:
Thanks for that!

Issue #137, opened 2019-07-13
Code style standardization


nemo1369:
Are there any code style standard the developer should keep close to?

I mean I'm going to bring a lot of changes in Boost.Multprecision, so we need to agree on some code style.

jzmaddock:
I suppose I should really write a .clang.format file for the library!

I'm not going to get on too much of a bikeshed discussion over style, the most important thing is to keep the code well commented and easy to read and understand.  I personally prefer well spaced out code to dense/terse formatting as you'll notice from the existing code, but otherwise I hope I'm not overly fussy.

Oh and make any PR's so I can see what's changed - which is to say please don't mix up formatting changes with substantive changes if you can help it.

What changes were you planning?  The next big change I want to get back to is full constexpr support - I did a fair bit of work on this last year (https://github.com/boostorg/multiprecision/pull/84) but the effort stalled due to a lack of `std::is_constant_evaluated`.  Now that we are starting to see support for that I need to revisit that..... I do need to finish off hypergeometrics for Boost.Math first though ;)

jzmaddock:
I've added a .clang-format file: it makes a lot of busy-body changes to existing code, but in the same spirit as what's there already (I hope!).

nemo1369:
That is totally great! I'm gonna automatically reformat the code to be compliant to the .clang_format introduced and come back with some pull request. It makes some minor changes to spacing inside various operator definitions, but this makes no big change.

nemo1369:
Coming to the full ```constexpr``` support, so I think, I would need to take this in sight, because my changes are all about underlying algorithms. I can make a monster-issue listing all the changes I want to bring, or I can create issue-by-issue along with pull requests with stuff I want to merge in. 

What would be the most convenient way to you?

jzmaddock:
If you can avoid monster PR's that would be appreciated - it makes it really hard to figure out what's going on ;)

nemo1369:
No, it is not about the monster PR, but about monster issue simply listing the upcoming changes.

nemo1369:
Since #142 is merged, I think this is not relevant anymore.

PR #140, opened 2019-07-14
Initial clang_format-based reformat #137

nemo1369:
Semi-automatic reformatting according to the ```.clang_format``` introduced.

This would allow us to adjust the ```.clang_format``` contents.

I think this is necessary because for further development I'm going to be using semi-automatic tools, so I think do you. This is the exact result it produced according to ```.clang_format```.

### DIFF
--- a/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
+++ b/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
@@ -6,67 +6,67 @@
 #ifndef BOOST_MATH_CONCEPTS_ER_HPP
 #define BOOST_MATH_CONCEPTS_ER_HPP
 
-#include <iostream>
-#include <sstream>
-#include <iomanip>
-#include <cmath>
+#include <boost/container_hash/hash.hpp>
 #include <boost/cstdint.hpp>
-#include <boost/multiprecision/number.hpp>
 #include <boost/math/special_functions/fpclassify.hpp>
 #include <boost/mpl/list.hpp>
-#include <boost/container_hash/hash.hpp>
+#include <boost/multiprecision/number.hpp>
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
 
-namespace boost{
-namespace multiprecision{
-namespace concepts{
+namespace boost {
+namespace multiprecision {
+namespace concepts {
 
 #ifdef BOOST_MSVC
 #pragma warning(push)
-#pragma warning(disable:4244)
+#pragma warning(disable : 4244)
 #endif
 
 struct number_backend_float_architype
 {
-   typedef mpl::list<boost::long_long_type>                 signed_types;
-   typedef mpl::list<boost::ulong_long_type>        unsigned_types;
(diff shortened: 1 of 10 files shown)


pabristow:
FWIW I note that the & and * are now part of the variable not the type

(number_backend_float_architype &o)
(number_backend_float_architype& o)

this feels C-ish to me.

Is this intended?  Just checking...

jzmaddock:
I confess it's not my preferred format, is there an option for this?

nemo1369:
Of course. Seems like your are looking for ```PointerAlignment: Left``` option.

Well, seems like we would need to unmerge this and I'ill reformat the code.

pabristow:
While I am bike-shedding about format :-) , I'm a strong fan of matching indenting of vertical brackets, despite the extra number of lines (though I'm not sure that John is).

int foo()
{
  return 42;
}
etc

And I trust we can still control some small bit of code's format with 
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code
?

I've always been sometimes cross with fully automatic stuff (though in general agreement).

jzmaddock:
I've corrected the pointer/reference issue in https://github.com/boostorg/multiprecision/pull/141.

nemo1369:
My proposal for code style standardization purpose is not to define the standard everyone is fine with, but to define the statements set, according to which contributors could write the code in any way they like, but call the clang-format in pre-commit hook would make it all the same.

For example, my personal code style is completely different from the one John is fine with, but I don't care, because at my side the code looks exactly as I want to. So I know, after the clang-format in pre-commit hook, it would look exactly like John wants in the repository.

@jzmaddock I got some reformatting done across the whole library according to the fixed .clang_format.

jzmaddock:
>While I am bike-shedding about format :-) , I'm a strong fan of matching indenting of vertical brackets, despite the extra number of lines (though I'm not sure that John is).

I'm fairly strongly in favour of the { aligning with the outer block and not being indented to match the content.

> And I trust we can still control some small bit of code's format with
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

Of course.

>I've always been sometimes cross with fully automatic stuff (though in general agreement).

There will always be WTF moments, but by and large it does a reasonable job.

> For example, my personal code style is completely different from the one John is fine with, but I don't care, because at my side the code looks exactly as I want to. So I know, after the clang-format in pre-commit hook, it would look exactly like John wants in the repository.

Nod, that should work fine, and having things locally as you're used to is always easier to grok.

>  I got some reformatting done across the whole library according to the fixed .clang_format.

I confess I was tempted, maybe I'll just do the lot and add it to the PR.

@pabristow  let me know if you spot any other SNAFU's.

jzmaddock:
> I confess I was tempted, maybe I'll just do the lot and add it to the PR.

Done and added to the PR.

jzmaddock:
Reverted and tried again with new .clang-format file: https://github.com/boostorg/multiprecision/pull/142

PR #141, opened 2019-07-15
Correct pointer and reference formatting in .clang-format.

jzmaddock:


pabristow:
Some spaces introduced are a bit/very odd.

For example

performance/cpp_bin_float_conversion_performance.cpp  line 24

- typedef int e_type;
+ typedef int                   e_type;

but several others that are even more odd.

:-(

NAThompson:
@jzmaddock : Do you want to put the same `.clang-format` into boost.math once you hammer out the style issues you dislike?

jzmaddock:
> performance/cpp_bin_float_conversion_performance.cpp line 24

If you look at the actual formatted file rather than the diff, it's not quite so bad:

```
   typedef int                   e_type;
   static boost::random::mt19937 gen;
   T                             val      = gen();
   T                             prev_val = -1;
```

This is because I turned on various alignment options:

```
AlignConsecutiveAssignments: true
AlignConsecutiveDeclarations: true
AlignEscapedNewlines: Left
AlignOperands: true
AlignTrailingComments: true
```

Which mostly produces rather nice results, but can make a bit of a mess of things if you have one typename that's much longer than the others.

Note that these options only effect consecutive statements, so for example in this situation, I added a couple of newlines, removed the clang-format introduced spacing and re-ran clang-format, and the result was way better.  In this situation, since we have 3 different things (A typedef, a static member, and 2 regular members, the extra newlines actually improve readability anyway IMO.  I don't propose to read through all the changes looking for others that could be improved, but it's something that could be corrected as and when maintenance is carried out on specific files (or if there are other shockers that are brought to my attention!)

jzmaddock:
Ooops, here's the improved version after adding newline:

```
template <class T>
T generate_random()
{
   typedef int e_type;

   static boost::random::mt19937 gen;

   T val      = gen();
   T prev_val = -1;

   while (val != prev_val)
   {
      val *= (gen.max)();
      prev_val = val;
      val += gen();
   }
   e_type e;
   val = frexp(val, &e);

   static boost::random::uniform_int_distribution<e_type> ui(-20, 20);
   return ldexp(val, ui(gen));
}
```

Will push shortly.

jzmaddock:
>Do you want to put the same .clang-format into boost.math once you hammer out the style issues you dislike?

I'm not sure, what do you think?

I do notice that all the regression tests are currently broken with this change - either clang-format is not as robust as we think or I've messed up somewhere.

jzmaddock:
The failures are caused by sorting of #includes, which for this library is a very bad idea, as many detail headers are not self-contained (nor are they intended to be), but rather part of a larger whole split up for readability.  I might have to close this and try again....

jzmaddock:
second try: https://github.com/boostorg/multiprecision/pull/142

NAThompson:
> I'm not sure, what do you think?

I've never been that picky about formatting. So for now, I guess it's probably not a priority.

pabristow:
The indenting is not a shocker, but I really don’t like it.

And the reason I think it is bad is that:

1 We DO have typenames that are widely varying in length, leading to some odd-looking long indents.  

2  Most important, it is like the spurious tidiness of left and right justification, despite research showing that ragged right is easier to read. The reason proposed is that the eye can use the ‘raggedness’ to keep focused on the line being read more easily.  And I think this is true.  

Aside I think that the same is true about comment alignment. 
Color is a much better way to separate comments (though tough on the chromatically challenged 😉 )
Keep things that are linked together, I say …

PR #142, opened 2019-07-15
Apply clang format

jzmaddock:


jzmaddock:
@nemo1369 I've hit a showstopper here: clang-format is mangling closing >'s in template declarations.  It's OK in C++11, but all the C++03 builds are broken.  This is despite putting

```
Standard: Cpp03
```

In the .clang-format file.  Any ideas anyone?

nemo1369:
@jzmaddock Is it ```SpaceAfterTemplateKeyword: false``` keyword you are looking for?

pabristow:
No -it is

"Standard (LanguageStandard)
Format compatible with this standard, e.g. use A<A<int> > instead of A<A<int>> for LS_Cpp03.

Possible values:

LS_Cpp03 (in configuration: Cpp03) Use C++03-compatible syntax."

Standard: Cpp03

that doesn't work for John.

Might it be 

Standard: LS_Cpp03 ??

(We are trying to avoid gratuitously making Boost.Math and Boost.Multiprecision no longer compatible with C++03.  Although some people are forging ahead to the latest version, there are platforms, compilers and peoples existing configuration that cannot move on easily, if at all.  So while it is fine to introduce C++11 and above in new code, we should not do this unnecessarily).

jzmaddock:
>Is it SpaceAfterTemplateKeyword: false keyword you are looking for?

No, it's changing some closing `> >` to `>>` even when the std version is set to 03.

nemo1369:
Okay. Since I've found  no configuration adjustment solution, I suggest to wrap the only macro definition where Clang.Format fails (in ```defaullt_ops.hpp```) with ```// clang-format off``` and ```// clang-format on``` and apply manual formatting.

Got it done in https://github.com/NilFoundation/multiprecision fork. I can make a new pull request.

PR #148, opened 2019-07-31
Correct docs for GCC's 'Q' suffix

jwakely:
There is no 'operator""Q', it's a built-in numeric suffix for floating-point literals, not a user-defined literal operator. Support for that suffix is disabled for -std=c++NN modes unless you use the -fext-numeric-literals option to enable the extension.

Also the correct spelling of the option is -std not --std.


jzmaddock:
Ah, I need to correct the source in multiprecision.qbk, I'll take care of that.

jwakely:
Oh sorry, I didn't realise that is a generated file.

pabristow:
Comments are still correct (and useful)!

PR #153, opened 2019-08-20
Fix Apple clang version detection false positives.

saurik:
In number.hpp, a workaround is presented for old versions of Apple clang. Apple uses non-standard version numbers, which makes it difficult to rely on ```__clang_major__```. The code first checked ```__APPLE_CC__``` in an attempt to check "is this compiler using Apple's version numbering", but open source upstream clang versions also define this variable when compiling to Apple targets; as this workaround breaks some client code, this makes some codebases impossible to compile for Apple targets with upstream clang. The correct variable to check is ```__apple_build_version__```.

### DIFF
--- a/include/boost/multiprecision/number.hpp
+++ b/include/boost/multiprecision/number.hpp
@@ -860,7 +860,7 @@ class number
       return this->template convert_to<T>();
    }
 #else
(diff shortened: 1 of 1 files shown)


saurik:
This bug is still here :(. I've updated my patch. Can this please get merged?

jzmaddock:
Thanks, CI failures are my fault (changes to Boost.Math).  Will merge now...

PR #143, opened 2019-07-15
Initial CMake build system implementation

nemo1369:
This makes the Boost.Multiprecision to support the CMake build system with following particular qualities:
- Build time increased up to 2 times.
- Some tests are temporarily disabled, they will be introduced a little bit later.
- Split to test suites in case of CMake-based build system usage is a little bit different for now. This means tests are being run by CTest, which is not configured for test suites support for now, but carefully annotates every test name, so it is still easy to know, what is wrong.
- CMake-based version uses standard ```find_package``` for Boost lookup, so it is required to install Boost to CI environment before it could be used by CMake. This needs a couple of improvements:
  - How to make ```bjam``` to build and install only particular libraries? 
  - Usage of BoostCMake (https://github.com/BoostCMake/boost_cmake) for Boost building is the key for the Boost installation avoidance. But the project linked is not really up to date for now. We will make it so, after that this project's CMake build system will be going to be updated as well as CI script.

PR #161, opened 2019-10-15
Modular adaptor

### LINKED ISSUE #144: Modular arithmetics adaptor
For several purposes it would be great for the Boost.Multiprecision to have a modular arithmetics adaptor.

Does it have any chance to be merged?

nemothenoone:
Initial modular adaptor implementation.

Consists of a couple of reduction techniques (Montgomery and Barrett reduction techniques). Several supplementary functions. Works for all the backends available now.

Still has a huge room for improvements (e.g. speed, docs etc) and this is coming.

Right now usage examples can be found in ```examples/modular_examples.cpp```.

Dependent on #143. Resolves #144 and https://github.com/NilFoundation/multiprecision/issues/2.


include/boost/multiprecision/modular/modular_adaptor.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+template <class Result, class Backend>
+inline void eval_convert_to(Result* result, const modular_adaptor<Backend>& val)
+{
+   using default_ops::eval_convert_to;
+   eval_convert_to(result, val.base_data());
+}
+
+template <class Backend, class T>
+inline typename enable_if<is_arithmetic<T>, bool>
```
jzmaddock: Formatting SNAFU? ;)

include/boost/multiprecision/cpp_modular.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <boost/multiprecision/cpp_int.hpp>
+
+namespace boost {
+namespace multiprecision {
+
+//mnt params
+typedef modular_params<cpp_int_backend<> > cpp_mod_params;
+
+// Fixed precision unsigned types:
+typedef modular_params<cpp_int_backend<128, 128, unsigned_magnitude, unchecked, void> >   umod_params_params128_t;
```
jzmaddock: Just a quick question - are fixed precision types appropriate here?  Which is to say do they still work correctly if the operation prior to the modular reduction would overflow?

test/test_modular_adaptor.cpp
```diff
@@ -0,0 +1,214 @@
...
+template <typename Backend>
+void run_tests()
+{
+   base_operators<Backend>();
+   comparsion_operators<Backend>();
+   bitwise_operators<Backend>();
+   pow_test<Backend>();
+}
+
+int main()
```
jzmaddock: Small, and rather pedantic point here: not withstanding the fact that I've written many tests like this, I've come to the conclusion that it's much more convenient to put the templates in a header and then write separate - and very short - .cpp files for each type tested.  It's just that it makes it easier to run and debug individual tests when you're not using the supplied build system (whether Boost.Build or CMake).

jzmaddock:
This is looking rather good - let me know when and what you want me to start looking through and testing stuff.

test/test_modular_adaptor.cpp
nemothenoone: Coming to tests.

I actually think most of tests should be rewritten to modern Boost.Test usage with all these ```BOOST_AUTO_TEST_SUITE``` and ```BOOST_FIXTURE_TEST_CASE```. So this particular solution you see is a really dumb one - we made it just to put some stub and make it work for now only.

include/boost/multiprecision/modular/modular_adaptor.hpp
nemothenoone: I've given up to clang-format. This is what it gave me. We can reformat this if required by disabling it with macro-attributes for these particular lines.

include/boost/multiprecision/cpp_modular.hpp
nemothenoone: That should work fine because modular params depend on internal limb storage size, which is non-mutable in case of fixed precision type. So, if the number was initialized properly with particular number, it should work fine.

Issue #167, opened 2019-11-19
str() gives weird result with global locale


dlaugt:
`str()` can give weird result mainly because it uses `boost::lexical_cast` and `stringstream` which are dependant of the global locale:
```cpp
locale::global (locale ("en-US"));
boost::multiprecision::cpp_dec_float_50 d {"1234.56"};
d.str ();
// gives 1,23.456,000,
```

jzmaddock:
Confirmed.

There are enough (lazy on my part) uses of lexical_cast that a larger solution is required - though I will no doubt pinch some of your suggested fix.

PR #168, opened 2019-11-19
Locale agnostic

dlaugt:
PR for #167


jzmaddock:
This should be fixed in develop now...

PR #179, opened 2020-01-02
bug in t1's size

madhur4127:
t1's size must be `2*n+2` because both t2 and t3 could be of size `n+1`. The test case that fails is square of `2^20000-1` because it causes resize (81 -> 82) and SIGABRT.

Also, made the `storage` tight.


madhur4127:
I think we should add this as a test (all 1s) because it is the worst case for the algorithm and it tests
- time
- space
- correctness

But please make sure that the exponent is big enough to make recursion deep.

jzmaddock:
Nod, we've both been looking at this I think.  I think this is now fixed (and tested) on this branch.

PR #186, opened 2020-01-24
Remove a couple of constexpr decorators, which are neither required n…

jzmaddock:
…or permitted


ckormanyos:
> Some checks were not successful

I feel as though the one single job that did not pass Travis CI build was a bad mojo event.
Line 211 of the faile build log says

```
GnuTLS recv error (-54): Error in the pull function.
```

Restart ehe CI build?

PR #189, opened 2020-01-29
Add better support for constexpr arithmetic in the next MSVC release.

jzmaddock:


pabristow:
Is this expected to work for recent Clang 9 that I think I remember has std:is_const_evaluated?

(Although I've stopped trying to use constexpr with cpp_bin_float as promote_args does not seem to be constexpr-able, so not trying this at present).

jzmaddock:
clang-9 was already supported.  This is for a yet-to-be-released MSVC.  cpp_bin_float currently has no constexpr support (I haven't got that far yet).

Issue #184, opened 2020-01-17
Feature request: quad-double precision.


cosurgi:
Hi,

as suggested in [math #303](https://github.com/boostorg/math/issues/303) I repost this here:

How feasible it is to integrate quad-double precision, It has 212 bits in significand and is faster than MPFR:
* [source code](https://www.davidhbailey.com/dhbsoftware/qd-2.3.22.tar.gz), it is on BSD license. 
* [article/documentation](https://www.davidhbailey.com/dhbpapers/quad-double.pdf)
* [quad double precision homepage]( https://www.davidhbailey.com/dhbsoftware/)

I have just recently [integrated boost multiprecision](https://gitlab.com/yade-dev/trunk/merge_requests/362) with a fairly large numerical computation software [YADE](https://yade-dem.org/doc/).

Float128 isn't enough for my needs, yade would greatly benefit from quad-double (as it is much faster than MPFR). Also it can serve as a really large testing ground. I have implemented [CGAL numerical traits](https://gitlab.com/yade-dev/trunk/blob/highPrecisionReal/lib/high-precision/CgalNumTraits.hpp) and [EIGEN numerical traits](https://gitlab.com/yade-dev/trunk/blob/highPrecisionReal/lib/high-precision/EigenNumTraits.hpp) to use multiprecision. And all of them are tested daily in our pipeline, see [an example run here](https://gitlab.com/yade-dev/trunk/pipelines/110095678).

cosurgi:
@ckormanyos 

>> How feasible it is to integrate quad-double precision...

> This is a good question. I would combine that with also software support for float128_t for those platforms lacking quadmath. In fact, I would see full cross-plattform availability of a float128_t (be it hardware backend or software pure) as a prerequisite before going for a float256_t.

This definitely would be great. Especially if float128 will use compiler/assembler version when available.

> Then there is the topic of design questions.
>
> Exrtract float128_t/256_t from the same template?
> Support other digit splits, etc.
> Lots of questions.

Yeah, it definitely depends on the approach. The simplest one is of course to use [qd-2.3.22.tar.gz](https://www.davidhbailey.com/dhbsoftware/qd-2.3.22.tar.gz) verbatim and only add the necessary multiprecision wrapper. A much more interesting approach is to extract the design used to "quadruple" the precision in there. In such a way that quad-float128 would be simply another template of the same design. 512bits is exactly what I will need in my calculations.

ckormanyos:
[...]

PR #190, opened 2020-01-29
Quad double: Add initial/tentative support for libqd.

jzmaddock:
See https://github.com/boostorg/multiprecision/issues/184.

### DIFF
--- a/include/boost/multiprecision/quad_double.hpp
+++ b/include/boost/multiprecision/quad_double.hpp
@@ -1,0 +1,660 @@
+///////////////////////////////////////////////////////////////
+//  Copyright 2012 John Maddock. Distributed under the Boost
+//  Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at https://www.boost.org/LICENSE_1_0.txt
+
+#ifndef BOOST_MATH_FLOAT_BACKEND_HPP
+#define BOOST_MATH_FLOAT_BACKEND_HPP
+
+#include <iostream>
+#include <iomanip>
+#include <sstream>
+#include <boost/cstdint.hpp>
+#include <boost/multiprecision/number.hpp>
+#include <boost/multiprecision/detail/float_string_cvt.hpp>
+#include <boost/container_hash/hash.hpp>
+#include <boost/type_traits/common_type.hpp>
+#include <boost/math/special_functions/trunc.hpp>
+#include <boost/math/special_functions/round.hpp>
+#include <qd/qd_real.h>
+
+namespace boost {
+namespace multiprecision {
+namespace backends {
+
+#ifdef BOOST_MSVC
+#pragma warning(push)
+#pragma warning(disable : 4389 4244 4018 4244 4127)
+#endif
+
+struct quad_double_backend
+{
+   typedef mpl::list<short, long long>                                 signed_types;
+   typedef mpl::list<unsigned short, unsigned long long>               unsigned_types;
+   typedef mpl::list<double, long double>                              float_types;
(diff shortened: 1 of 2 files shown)


ckormanyos:
> See #184.

Awesome ! I was thinking you might start moving forward on this. Looks like a promising start...

ckormanyos:
I would like to refresh interest in this project (and can mentor some additional potential progress on it) for [Boost Google Summer of Code 2021](https://github.com/boostorg/wiki/wiki/Google-Summer-of-Code:-2021)

jzmaddock:
SOC is a good thought, the code that's here works, needs more testing (there's one unresolved issue I think), and if possible it would be great to move this to C++ only (ie without the external library).

ckormanyos:
> SOC is a good thought, the code that's here works, needs more testing (there's one unresolved issue I think), and if possible it would be great to move this to C++ only (ie without the external library).

Agreed... on all points. I will comply in that order.

ckormanyos:
I tried a big merge operation of today's develop onto this relatively old branch.

ckormanyos:
Develop has been successfully merged into this branch. This is good and partially prepares the road for further work on this topic. If (or when) I personally do something on QD, I think we should simply use (re-use, continue to use) this PR.

Is that OK to keep going on QD on this PR, at least at first for a while?

jzmaddock:
Good question, I guess if it's a GSOC project then it might be easier to fork this off into the student's own repro?

AuroraDysis:
I tried to use qd in my program several times, but according to my experience, it has a lot of bugs. I dare say that I am not alone and there are others who experienced a similar situation. For example

https://github.com/ddmarshall/Code-Eli/issues/27#issue-23182272

ckormanyos:
> tried to use qd in my program several times, but according to my experience, it has a lot of bugs. 

The intent of integrating this QD thing is to either use it (if it's good enough out of the box) or replacing entirely its algorithms under Boost licensing if it's not. At the moment, we have not fully or exhaustively estimated the work/effort in these activities, but the potential GSoC is expected to move this endeavor forward.

anilmuthigi:
I was going through this function : 

inline void eval_divide(quad_double_backend& result, const quad_double_backend& a, const quad_double_backend& b)
 {
    bool i1 = (boost::math::isinf)(a.data()[0]);
    bool i2 = (boost::math::isinf)(b.data()[0]);
    if (i1 && !i2)
       result.data() = a.data();
    else if (i2 && !i1)
       result.data() = 0.0;
    else
       result.data() = a.data() / b.data();
 }
 
 Don't you think that b.data()=0.0 should be handled separately?

jzmaddock:
>Don't you think that b.data()=0.0 should be handled separately?

I guess only if the division doesn't generate an infinity anyway - and I'm assuming it does?

ckormanyos:
> going through this function

Yes. Thanks for looking.

There are a whole bunch of functions in this kind of multiprecision adaption. As this project moves on, we will be looking into this function and numerous others, re-writing some, adaptimg others, all with the common goal of getting an efficient, portable, reliable, tested QD running.

I think we wil luncover a bunch of little nuances like this one...

jzmaddock:
>I think we wil luncover a bunch of little nuances like this one...

Indeed, what I should have said was: if the special case is already handled by the hardware, then we should leave it to it on the grounds that it will be way more efficient than any code we can write.

ckormanyos:
> if the special case is already handled by the hardware

Yes John. Agreed. That's actually a detail that I'll be consulting you on regarding advice and best way to/or not to do it, later in the project.

PR #193, opened 2020-02-07
Documentation update

jzmaddock:
Made a PR so we can discuss the changes more easily.


config/Jamfile.v2
```diff
@@ -34,6 +41,7 @@ project : requirements
    <toolset>msvc:<warnings>all
    <toolset>gcc:<cxxflags>-Wall
    <toolset>gcc:<cxxflags>-Wextra
+   <toolset>gcc:<cxxflags>-fext-numeric-literals  # enable suffix Q for float128.
```
jzmaddock: No, we *DO NOT* support -fext-numeric-literals only -std=gnu++XX - this puts gcc into an odd state where neither ourselves nor the C++ std lib can determine whether `__float128` is fully supported or not.  Plus the whole point of the configuration checks is to determine whether or not __float128 is fully supported -  we shouldn't be second guessing this, just determine whether the current compilation mode is OK or not.

config/has_constexpr_limits_cmd.cpp
```diff
@@ -3,11 +3,13 @@
...
 
+// This determines if std::numeric_limits can be constexpr for the compiler/configuration.
+
 #ifndef __GNUC__
-#error "Compiler is not GCC"
+#  error "Compiler is not GCC."
 #endif
 #if __GNUC__ < 9
-#error "Older GCC versions don't support -fconstexpr-ops-limit"
+#  error "Older GCC versions don't support -fconstexpr-ops-limit (aka -fconstexpr-loop-limit=n) ."
```
jzmaddock: I don't think these are the same at all?  -fconstexpr-loop-limit deals with loops within constexpr functions, where as -fconstexpr-ops-limits sets the total number of integer operations permitted.

config/has_float128.cpp
```diff
@@ -3,10 +3,12 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+// This program tests if a configuration can use type __float128.
+
 #include <boost/config.hpp>
 
 #ifndef BOOST_HAS_FLOAT128
-#error "This doesn't work unless Boost.Config enables __float128 support"
+#  error "This doesn't work unless Boost.Config enables __float128 support. Requires GCC using GNU C++compiler and option -fext-numeric-literals. For example: b2/bjam command "b2 float128_snips toolset=gcc-8.1.0 cxxstd=17 cxxstd-dialect=gnu address-model=64 release "
```
jzmaddock: As above, -fext-numeric-literals is not supported, only -std=gnu++XX

config/has_is_constant_evaluated.cpp
```diff
@@ -3,8 +3,44 @@
 //  Boost Software License, Version 1.0. (See accompanying file
 //  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 
+// This program determines if std::is_constant_evaluated  is available to switch within a function to use compile-time constexp calculations.
```
jzmaddock: I'm not sure about these changes - the purpose of this configuration test is to determine whether BOOST_MP_HAS_IS_CONSTANT_EVALUATED has been set or not.

* There are some compilers (gcc-6/7/8) which do support constexpr detection through means other than `std::is_constant_evaluated`, and in compilation modes other than C++2a.
* Likewise `__builtin_is_constant_evaluated` is available in C++14 and later once it's been added, so some of these checks just create unnecessary warnings?
* There are some compilers (Intel - support for which I'm still working on), which does support `__builtin_is_constant_evaluated` but which is too buggy to use it's C++14 constexpr support at all.  So again IMO this just generates unnecessary warning noise?

doc/html/standalone_HTML.manifest
```diff
@@ -0,0 +1,88 @@
+index.html
```
jzmaddock: Did you mean to commit this one - these manifests are of no use to anyone that I can see?

Not sure why they're generated at all to be honest... it's a docbook thing :(

example/Jamfile.v2
```diff
@@ -60,7 +60,19 @@ project
       <toolset>msvc:<cxxflags>/wd4701
       <toolset>msvc:<cxxflags>/wd4127
       <toolset>msvc:<cxxflags>/wd4305
-    ;
+      <toolset>clang-win:<link>static  # Clang-win does not generate .dlls.
+      <toolset>clang:<link>static # Clang-linux does not generate .dlls.
```
jzmaddock: Caution: this is only on windows, not Linux.

example/big_seventh.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+To get as precise-as-possible decimal fractions like 1.234, we can write
+*/
+
+  const cpp_bin_float_50 f1(cpp_bin_float_50(1234) / 100);  // Construct from a fraction.
+  std::cout << "cpp_bin_float_50 f1(cpp_bin_float_50(1234) / 100) = " << f1 << std::endl; // 12.340000000000000000000000000000000000000000000000
+/*`or
+*/
+  const cpp_bin_float_50 f2("1.234");  // Construct from decimal digit string.
+  std::cout << "cpp_bin_float_50 f2(\"1.234\") = " << f2 << std::endl; // 12.340000000000000000000000000000000000000000000000
```
jzmaddock: Oops, factor of 10 out?

example/float128_snips.cpp
```diff
@@ -3,40 +3,91 @@
...
-   //
-   // We can declare constants using GCC or Intel's native types, and the Q suffix,
-   // these can be declared constexpr if required:
+   /*` We can declare constants using GCC or Intel's native types, and literals with the Q suffix, and these can be declared `constexpr` if required:
+   */
    /*<-*/
 #ifndef BOOST_NO_CXX11_CONSTEXPR
    /*->*/
-   constexpr float128 pi = 3.1415926535897932384626433832795028841971693993751058Q;
+   constexpr float128 pi = 3.14159265358979323846264338327950Q;
```
jzmaddock: That doesn't seem to have enough digits anymore?

example/gauss_laguerre_quadrature.cpp
```diff
@@ -225,7 +225,7 @@ class guass_laguerre_abscissas_and_weights : private boost::noncopyable
         // bisection steps in order to tighten up the root bracket.
         boost::uintmax_t a_couple_of_iterations = 3U;
         const std::pair<T, T>
-          first_laguerre_root = boost::math::tools::bisect(laguerre_object,
+          first_laguerre_roots = boost::math::tools::bisect(laguerre_object,
```
jzmaddock: @NAThompson : `first_laguerre_root`(s) appears to be unused?

There also seem to be some branches in the logic which do not set `first_laguerre_root_has_been_found`?

test/Jamfile.v2
```diff
@@ -58,8 +58,11 @@ project : requirements
    <toolset>msvc:<cxxflags>/fp\:precise
    <toolset>intel-win:<runtime-link>static
    <toolset>intel-win:<link>static
-   <toolset>clang-win:<link>static
-   
+   <toolset>clang-win:<link>static  # Clang-win does not generate .dlls.
+   <toolset>clang:<link>static # Clang-linux does not generate .dlls.
```
jzmaddock: As above this does not apply to non-windows platforms.

example/gauss_laguerre_quadrature.cpp
NAThompson: I'm baffled! I'm pretty sure I didn't write any of this, but there's my name in the copyright! I read the `git blame`; looks like @ckormanyos write it.

config/has_float128.cpp
pabristow: Changed to `#  error "Older GCC versions don't support -fconstexpr-ops-limit (see also -fconstexpr-loop-limit=n) ."
`

doc/html/standalone_HTML.manifest
pabristow: I just committed all the Quickbook generated files.  I can't see any particular use for the manifest.

I've Perhaps this should be added to the .gitignore?

I would also favour not committing the entire generated html and adding the /html folder to the 'gitignore.  I can see that it is convenient for other developers (or even users) who don't have to rebuild the docs.  But it is a PITA to have all these files that keep changing (and must be cluttering up GIThub).  Would generating a PDF version for 'users' be more convenient?  (It has the side effect of being more easily searchable: unless I know the right section, I haven't found how to search the html folder easily.  Once 'published' Google is our friend.

example/big_seventh.cpp
pabristow: Oops indeed!

example/gauss_laguerre_quadrature.cpp
pabristow: I'll try to establish the true author, and absolve you of all possible blame for any hanging branches!

ckormanyos: > looks like @ckormanyos ... [wrote] it

I did. I will look at the example and see if the variable is used or not.
I will also check that example.

pabristow:
I’m still getting my head right around this example.

I concluded that the warnings were correct and that a mid-point of the first estimate was unassigned, as you suggest.

But the range of alpha was not used, so it didn’t make any difference.

I note that the BOOST_CONSTEXPR, BOOST_STATIC_CONSTEXPR, BOOST_CONSTEXPR_OR_CONST  MACROS could be used?

https://www.boost.org/doc/libs/1_72_0/libs/config/doc/html/boost_config/boost_macro_reference.html

This might nicely exercise constexpr with multiprecision?

It would be nice to convert to Quickbook snippets (and to reference from Boost.Math docs as well as a good example of using root finding.

Tell what you would like me to do.

ckormanyos:
> I will look at the example and see if the variable is used or not.

I am at the moment quite busy. I will attempt to check out the example on the weekend.
Best regards, Chris

example/float128_snips.cpp
pabristow: I reduced this to 36 decimal digits to match the number with the implied precision of float128 of std::numeric_limits::<float128>max_digits10 = 36 as I felt this was clearer.  But I could revert it if you don't think that this helps.

jzmaddock: There aren't 36 digits?

pabristow: You are, as always, this is digits10 rather than max_digits10.  I'll change it.

example/gauss_laguerre_quadrature.cpp
ckormanyos: >>> There also seem to be some branches in the logic which do not set `first_laguerre_root_has_been_found`?

Correct. Thanks. Good catch. I have refactored these issues.

>>I'll try to establish the true author

> I will look at the example and see if the variable is used or not.

I have refactored the example Gauss Laguerre quadrature
[here](http://github.com/boostorg/multiprecision/tree/gauss_laguerre)

ckormanyos: See #196
We can merge those changes to this PR if desired.

doc/multiprecision.qbk
```diff
@@ -1091,57 +1108,85 @@ the default by defining either `BOOST_MP_USE_FLOAT128` or `BOOST_MP_USE_QUAD`.
 * When the underlying type is Intel's `_Quad` type, the code must be compiled with the compiler option `-Qoption,cpp,--extended_float_type`.
 * When compiling with `gcc`, you need to use the flag `--std=gnu++11/14/17`, as the suffix 'Q' is a GNU extension. Compilation fails with the flag `--std=c++11/14/17`
 unless you also use `-fext-numeric-literals`.
+* You will need to link to `libquadmath.dll` with the link command `-lquadmath` and ensure that the DLL is visible by the linker.
```
jzmaddock: I worry this is too windows centric?

doc/multiprecision.qbk
```diff
@@ -1091,57 +1108,85 @@ the default by defining either `BOOST_MP_USE_FLOAT128` or `BOOST_MP_USE_QUAD`.
 * When the underlying type is Intel's `_Quad` type, the code must be compiled with the compiler option `-Qoption,cpp,--extended_float_type`.
 * When compiling with `gcc`, you need to use the flag `--std=gnu++11/14/17`, as the suffix 'Q' is a GNU extension. Compilation fails with the flag `--std=c++11/14/17`
 unless you also use `-fext-numeric-literals`.
+* You will need to link to `libquadmath.dll` with the link command `-lquadmath` and ensure that the DLL is visible by the linker. 
+If you are using the B2/bjam build system then commands`<linkflags>-lQUADMATH` and `<linkflags>-L"path/to/lib"` will be needed.
```
jzmaddock: QUADMATH in upper case will *not* work on Linux, it should be all lower case on both systems.

doc/multiprecision.qbk
```diff
@@ -1091,57 +1108,85 @@ the default by defining either `BOOST_MP_USE_FLOAT128` or `BOOST_MP_USE_QUAD`.
...
+* The values shown by `std::numeric_limits<float128>` and extremely close ['but not identical]
+to those from the equivalent precision and range multiprecision types `std::numeric_limits<cpp_bin_float_quad>` and `std::numeric_limits<cpp_dec_float_quad>`.
```
jzmaddock: Actually, I hope they should be exactly the same, I haven't checked though have you?

doc/multiprecision.qbk
```diff
@@ -1175,7 +1220,7 @@ much greater precision and implementing interval arithmetic.
 Type `mpfi_float_backend` can be used at fixed precision by specifying a non-zero `Digits10` template parameter, or
 at variable precision by setting the template argument to zero.  The `typedef`s `mpfi_float_50`, `mpfi_float_100`,
 `mpfi_float_500`, `mpfi_float_1000` provide arithmetic types at 50, 100, 500 and 1000 decimal digits precision
-respectively.  The `typedef mpfi_float` provides a variable precision type whose precision can be controlled via the
+respectively.  The `typedef mpfi_float` provides a variable precision type whose precision can be controlled via theF
```
jzmaddock: Fat fingered 'F' ? ;)

doc/multiprecision.qbk
```diff
@@ -2149,15 +2194,20 @@ or GCC-6 or later in C++14 mode.
...
 [constexpr_circle]
 
-We can now calculate areas and circumferences using all constexpr arithmetic:
+We can now calculate areas and circumferences, using all compile-time `constexpr` arithmetic:
 
 [constexpr_circle_usage]
 
-Note that these make use of the numeric constants from the Math library, which also happen to be `constexpr`.
+Note that these make use of the numeric constants from the __math_constants library, which also happen to be `constexpr`. 
+These usually have the full precision of the floating-point type, here 128-bit, about 36 decimal digits.
```
jzmaddock: Hopefully, they should always have the full precision of the FT type?

example/big_seventh.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+   cpp_bin_float_50 seventh_double = cpp_bin_float_50(1./7);  // Construct from double! (0.14285714285714)
+   std::cout << "seventh_double = "  << seventh_double << std::endl; // Boost.Multiprecision post-school error!
+   // seventh_double = 0.14285714285714284921269268124888185411691665649414
+
+/*`Did you spot the probably-unintended difference?  After the 17th decimal digit, result is apparently random
+and not the expected recurring pattern 14285714285714...
+
+The 'random' digits after digit 17 are from the `cpp_bin_float_50` representation of the `double` value 0.14285714285714
+which is different from the 'better' `cpp_bin_float_50` representation of the fraction 1/7
```
jzmaddock: This is a bit clunky but I don't know how to re-phrase - the cpp_bin_float_50 represents the double precision value exactly, but of course all the extra bits are set to zero.  The apparently "random" digits in the decimal representation aren't random at all - they're an artefact of the base-2 to 10 conversion (ie we can never represent the binary value exactly in base 10, so the decimal digits go on and on forever).  Apologies for being a pedant!

example/big_seventh.cpp
```diff
@@ -0,0 +1,181 @@
...
+*/
+
+  const cpp_bin_float_50 f1(cpp_bin_float_50(1234) / 100);  // Construct from a fraction.
+  std::cout << "cpp_bin_float_50 f1(cpp_bin_float_50(1234) / 100) = " << f1 << std::endl; // 12.340000000000000000000000000000000000000000000000
+/*`or
+*/
+  const cpp_bin_float_50 f2("1.234");  // Construct from decimal digit string.
+  std::cout << "cpp_bin_float_50 f2(\"1.234\") = " << f2 << std::endl; // 12.340000000000000000000000000000000000000000000000
+
+/*`that are different from constructing from a `double` with value 1.234
```
jzmaddock: ... "since any binary floating point type only represents decimal fractions approximately."

doc/multiprecision.qbk
ckormanyos: > I worry this is too windows centric?

As it stands, it's not quite correct because you will not be linking to the DLL.

One option would be to use a more vague, yet very simplified generic form:

- You need to link with `quadmath` and also ensure that symbols from the quadmath library are available.

Another option would be to describe more precisely for Win* or *nix separately.

- On `*nix` systems, you need to link with `quadmath` by providing the linker flag `-lquadmath`, which will link with `libquadmath.a`.
- On `Win*` systems such as GCC ported to MinGW, you also need to provide the linker with the `-lquadmath` flag in order to link with `libquadmath.lib`. It might also be necessary to ensure that symbols from the quadmath DLL are available at runtime.

jzmaddock: Perfect!  Glad one of us has a brain this evening! :)

doc/multiprecision.qbk
pabristow: Corrected to lower case.

doc/multiprecision.qbk
pabristow: Fixed (but not the Fat Fingers :-( ).

doc/multiprecision.qbk
pabristow: These should always have the full precision of the floating-point type, here 128-bit, about 36 decimal digits.

example/big_seventh.cpp
pabristow: Is this any better (if even clunkier?)   

The 'apparently random' digits after digit 17 are from the `cpp_bin_float_50` representation 
of the `double` value 0.14285714285714 that is, in effect, converted 
from 0.1428571428571400000000000000000000000000000000000
which is different from the 'nearest representable' or 'best' `cpp_bin_float_50` representation of the fraction 1/7

PS "The term, then, is obviously a relative one: my pedantry is your scholarship, his reasonable accuracy, her irreducible minimum of education and someone else’s ignorance."―H. W. Fowler, Modern English Usage.

doc/multiprecision.qbk
pabristow: When I previously checked a few, some appeared a bit different in the 'noisy' digits.  I have rechecked and now understand that this was an aberration caused by a previous difference in max_digits10 (now both max_digits10=36) and one display showing trailing zeros, and the other not.  (support for std::hex would help with this).

Now looks exact, but I haven't done an exhaustive check.   (should I) So now changed to:

* The values shown by `std::numeric_limits<float128>` are the same as those from the equivalent precision and range multiprecision types `std::numeric_limits<cpp_bin_float_quad>` and `std::numeric_limits<cpp_dec_float_quad>`.

I also propose to add an example program and include a listing all the numeric limits values for a variety of types, including these of course.  I feel users will find these convenient.

jzmaddock:
Apologies @pabristow ball dropping again, is there anything more to do here?

pabristow:
No - you haven't dropped the ball - I'm holding it up!

I believe I have dealt with all your comments.

I have this on hold waiting to merge with some update to add the Gauss-Lagrange example before merging the whole docs-update into Multiprecision.  I'll check out what Chris has written or agree something about it.

ckormanyos:
>  I'll check out what Chris has written...

The Gauss-Laguerre example benefited from #193 

Details we could work into the docs can be found in [lines 8 to 30 here](https://github.com/boostorg/multiprecision/blob/d2182773ac356de6e3a5e70afbc8f7bc48e33efa/example/gauss_laguerre_quadrature.cpp#L8)

ckormanyos:
> I'll check out what Chris has written

The Gauss-Laguerre example is now better thanks to this PR.

More details we can work into docs can be found in [lines 8 to 30 here](https://github.com/boostorg/multiprecision/blob/d2182773ac356de6e3a5e70afbc8f7bc48e33efa/example/gauss_laguerre_quadrature.cpp#L8)

ckormanyos:
> The Gauss-Laguerre example is now better thanks to this PR.

I have locally cherry picked Gauss-Laguerre example. Can commit any time if you would like me to.

Would you like me to merge and push the improved Gauss-Laguerre example cpp file from my branch into into your branch?

Kind regards, Chris

pabristow:
Yes - please do that.  Then I will use the text you suggest to add to the .qbk, links to the example etc, and recheck.  I'll then merge docs updates into develop.

ckormanyos:
> Yes - please do that. Then I will use the text you suggest to add to the .qbk, links to the example etc, and recheck. I'll then merge docs updates into develop.

Done. Thank you.
Please continue with docs.

Comments are scattered throughout the example, some are potentially fit for doc purposes.

Especially [here](https://github.com/boostorg/multiprecision/blob/28c16704be1f6e558c4177abb5b7cf26bac644fc/example/gauss_laguerre_quadrature.cpp#L8)

and [here](https://github.com/boostorg/multiprecision/blob/28c16704be1f6e558c4177abb5b7cf26bac644fc/example/gauss_laguerre_quadrature.cpp#L480)

ckormanyos:
> I'll work on addition to docs

Great. Ready when you are, Paul.

madhur4127:
Here's from [TODO](https://www.boost.org/doc/libs/1_71_0/libs/multiprecision/doc/html/boost_multiprecision/map/todo.html):

> Add better multiplication routines (Karatsuba, FFT etc) to cpp_int_backend.

Karatsuba is done and I think we discussed that FFT would not be beneficial as it has crossover point ~million limbs.

> Add assembly level routines to cpp_int_backend.

It seems we've rejected this proposal.

@jzmaddock, @ckormanyos should I add the things that John pointed out like modular exponentiation, sqrt, gcd to this? Or we can use Github's Project as Mateusz Loskot does for [GIL](https://github.com/boostorg/gil/projects)

ckormanyos:
To your first points, Madhur...

>> Add better multiplication routines (Karatsuba, FFT etc) to cpp_int_backend.

> Karatsuba is done and I think we discussed that FFT would not be beneficial as it has crossover point ~million limbs.

Actually, I would still like to mention these in the following form:
Add better multiplication routines (Toom-Cook, FFT) to cpp_int_backend.

>> Add assembly level routines to cpp_int_backend.

> It seems we've rejected this proposal.

Correct.

ckormanyos:
>  Then I will use the text you suggest to add to the .qbk, links to the example etc, and recheck. I'll then merge docs updates into develop.

The part on Gauss-Laguerre example looks good to me. Thank you.

PR #196, opened 2020-02-16
Gauss laguerre

ckormanyos:
Remove unused variables.
Check with GCC's -Wall -Wextra -pedantic and a few more warnings.
Remove all MSVC level 3 warnings.
Other compiler warnings not checked.
Eliminate unused logic paths.
Speedup and calibrate for 50...1200 decimal digits.
Use a few more Boost macros, but not yet fully constexpr.
Repair the correct author and update the copyright information.

### DIFF
--- a/example/gauss_laguerre_quadrature.cpp
+++ b/example/gauss_laguerre_quadrature.cpp
@@ -467,16 +467,17 @@ namespace detail
 
 } } // namespace gauss::laguerre
 
-namespace local
+struct local
 {
-  BOOST_CONSTEXPR unsigned int my_digits10 = 101U;
+  BOOST_STATIC_CONSTEXPR unsigned int my_digits10 = 101U;
 
-  using float_type =
-    boost::multiprecision::number<boost::multiprecision::cpp_dec_float<local::my_digits10>,
-                                  boost::multiprecision::et_off>;
-}
+  typedef boost::multiprecision::number<boost::multiprecision::cpp_dec_float<my_digits10>,
+                                        boost::multiprecision::et_off>
+  float_type;
+};
 
-BOOST_STATIC_ASSERT_MSG(local::my_digits10 > 20U, "Error: This example is intended to have more than 20 decimal digits");
+BOOST_STATIC_ASSERT_MSG(local::my_digits10 > 20U,
+                        "Error: This example is intended to have more than 20 decimal digits");
 
 int main()
 {
@@ -506,7 +507,7 @@ int main()
 
   std::cout << "laguerre_order: " << laguerre_order << std::endl;
 
-  using abscissas_and_weights_type = gauss::laguerre::detail::abscissas_and_weights<local::float_type>;
+  typedef gauss::laguerre::detail::abscissas_and_weights<local::float_type> abscissas_and_weights_type;
 
   const abscissas_and_weights_type the_abscissas_and_weights(laguerre_order, local::float_type(-1) / 6);
 
--- a/example/gauss_laguerre_quadrature.cpp
+++ b/example/gauss_laguerre_quadrature.cpp
@@ -5,6 +5,31 @@
 //          http://www.boost.org/LICENSE_1_0.txt)
 //
 
+// This example uses Boost.Multiprecision to implement
+// a high-precision Gauss-Laguerre quadrature integration.
+// The quadrature is used to calculate the airy_ai(x) function
+// for real-valued x on the positive axis with x.ge.1.
+
+// In this way, the integral representation could be seen
+// as part of a scheme to calculate real-valued Airy functions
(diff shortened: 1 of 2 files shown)


ckormanyos:
This PR is intended to implement the first round or repairs and improvements of the Gauss Laguerre quadrature example from 2012-2015.

Some of these points have been mentioned in #193

ckormanyos:
Actually, I would like to extend this example.
Will close this PR without merging it.

ckormanyos:
The Gauss-Laguerre quadrature example has been reworked.
The approximation now handles arguments on the real axis airy_ai(x) with x.gt.1.
The example now checks a range of arguments with control values from `cyl_bessel_k` of 1/3.
The idea of this example is now much more clear, as mentioned in code comments.

- This example is intended to handle airy_ai(x) calculations for medium to large arguments
- Use quadrature for x >=1
- Assume that and a hypergeometric function or a Taylor series could be used for x < 1.

I will probably merge my own PR in a day or so, assuming the tests run.

NAThompson:
Very cool. Just a few comments: I added a progress bar into:

https://github.com/boostorg/math/blob/develop/example/naive_monte_carlo_example.cpp

It seems like this example might also benefit from the `display_progress` function, since it takes a while to run, with no output for some time, and then lots of output which could potentially be better expressed as a progress bar.

I ran this example through the Sanitizer Suite over both clang++-9 and g++-7; it came out clean. So looks good to me!

example/gauss_laguerre_quadrature.cpp
```diff
@@ -26,258 +31,222 @@
...
-
-        j_alpha_m1 = alpha * (((((                             + 0.043F
-                                  * alpha_pow_minus_two_thirds - 0.0908F)
-                                  * alpha_pow_minus_two_thirds - 0.00397F)
-                                  * alpha_pow_minus_two_thirds + 1.033150F)
-                                  * alpha_pow_minus_two_thirds + 1.8557571F)
-                                  * alpha_pow_minus_two_thirds + 1.0F);
+        first_laguerre_root = (j_alpha_m1_sqr * (   -T(0.6666666666667F)
+                                                 + ((T(0.6666666666667F) * alpha) * alpha)
+                                                 +  (T(0.3333333333333F) * j_alpha_m1_sqr) + vf2)) / (vf2 * vf);
```
NAThompson: Doesn't this lose precision? I might do T(2)/T(3), or boost::math::constants::two_thirds.

pabristow:
I pondered on this too, but concluded that it is only a guess, so double is more than good enough?

Looks good (if it took a little while for a superannuated FORTRAN programmer to et his head around some of the lambda-ing).  While this is indubitably cool, is there a performance reason for doing it this way?

NAThompson:
Whoops, didn't notice that it's a guess; my bad.

ckormanyos:
> It seems like this example might also benefit from the display_progress function, since it takes a while to run, with no output for some time, and then lots of output which could potentially be better expressed as a progress bar.

Kewl. I will check it out...

I think this example is progressing to be about where I would like it to be.

Based on some new knowledge that I acquired in this area and extension of the example, I think it will make sense that I add some further sentences and an equation or two to the docs, as time permits.

Thanks for the help, and kind regards, Chris

ckormanyos:
Is there a simple reason anyone knows why the builds are failing on CI?

I can't see how my changes caused these failures. But I am still investigating...

ckormanyos:
>> It seems like this example might also benefit from the display_progress function, since it takes a while to run, with no output for some time, and then lots of output which could potentially be better expressed as a progress bar.

> Kewl. I will check it out...

Progress percent indications have been added.

NAThompson:
@ckormanyos : Just tried it out; looks great!

ckormanyos:
> Testing especially looking much better.

Yup. Thank you. It was a good idea from your side to clean up this example. I learned a lot and I think the example is much better now.

> I noted this remaining comment...

Which is now expressed differently, thereby removing the open endedness of the comment.

More detailed comments describing the example are also now near the head of the file.

> As far as I can see all the results are 'exact', so perhaps the guard digits are a little over-enthusiastic?

I have removed the guard digits. It was necessary to slightly increase the tolerance of close fraction comparison with the control values. But I do, in fact, prefer this now without the guard digits. It's a bit cleaner to describe and comprehend.

The tests are now cycling. I had a few problems with using some C++11 stuff on the appveyor runs. But I switched over to the Boost macros such as those for older compilers lacking some C++11 features.  Barring any unexpected failures, I will handle this PR myself after the tests cycle.

Thanks and kind regards, Chris

ckormanyos:
The Gauss-Laguerre quadrature example has been refactored and improved.

PR #199, opened 2020-03-08
Update tommath.hpp to cope with recent tommath releases.

jzmaddock:
Replaces https://github.com/boostorg/multiprecision/pull/181.


madhur4127:
Thanks John,

Somehow my github dashboard didn't show #181 but only #182 - Comba multiplier. So I totally forgot about the PR :(

I will try to be careful next time, I will close #181.

PR #200, opened 2020-03-10
Mark C++03 as deprecated.

jzmaddock:


madhur4127:
I believe we can make the codebase more expressive and Modern C++ish thereby enhancing readability perhaps efficiency too!

jzmaddock:
> Have I misunderstood this, or should the title really be 'C++03 deprecated'?

Sorry my bad - brain cells are in short supply at present while I battle "Ye Olde Man Flu" (not yet the newsworthy variety I hasten to add! :) )

So yes C++03 is deprecated, and C++11 is required.  And yes the length of the list surprised me considering this was conceived as a C++98 library!

pabristow:
Sympathy - go back to bed? ;-)

Issue #207, opened 2020-03-30
mpfr.hpp math log(…) function conflict with boost::log; compilation error while invoking log(…)


cosurgi:
This line won't compile:

https://github.com/boostorg/multiprecision/blob/b46a0ecbe5301858ef9016ea6e2e92e2afc399c5/include/boost/multiprecision/mpfr.hpp#L2246

My guess is that only `eval_log(…)` is available there. I encountered this problem in yade pipeline with boost 1.71: https://gitlab.com/yade-dev/trunk/-/jobs/491138134

```
/usr/include/boost/multiprecision/mpfr.hpp: In function 'boost::multiprecision::number<boost::multiprecision::backends::mpfr_float_backend<digits10, AllocationType>, ExpressionTemplates> boost::math::lgamma(boost::multiprecision::number<boost::multiprecision::backends::mpfr_float_backend<digits10, AllocationType>, ExpressionTemplates>, int*, const Policy&)':
 /usr/include/boost/multiprecision/mpfr.hpp:2233:22: error: expected primary-expression before '(' token
  2233 |          result = log(boost::math::constants::pi<boost::multiprecision::number<boost::multiprecision::mpfr_float_backend<Digits10, AllocateType>, ExpressionTemplates> >()) - lgamma(arg, 0, pol) - log(t);
       |                      ^
 /usr/include/boost/multiprecision/mpfr.hpp:2233:200: error: expected primary-expression before '(' token
  2233 |          result = log(boost::math::constants::pi<boost::multiprecision::number<boost::multiprecision::mpfr_float_backend<Digits10, AllocateType>, ExpressionTemplates> >()) - lgamma(arg, 0, pol) - log(t);
       |                                                                                                                                                                                                        ^
```


~~I think that replacing it with `eval_log(…)` fixes the problem~~. At least this works for me, and test invocations of lgamma pass `yade --test`. I will prepare a PR with a fix.

EDIT: adding `using boost::multiprecision::log;` fixes conflict with `boost::log` namespace.

cosurgi:
Also see: https://gitlab.com/yade-dev/trunk/-/merge_requests/450

cosurgi:
Boost 1.71 is about to be included in Ubuntu LTS 20.04. Without the fix for this yade using MPFR will be impossible for next 5 years on ubuntu. I understand it's late, but there is still some chance that this bugfix will enter ubuntu 20.04 if you accept it here.

I tried to implement a workaround for that, but without success so far.

[...]

PR #208, opened 2020-03-30
Fix https://github.com/boostorg/multiprecision/issues/207

cosurgi:
See https://github.com/boostorg/multiprecision/issues/207


jzmaddock:
Not the right fix.

PR #209, opened 2020-03-31
Make calls to log qualified so that they don't get poisoned by Boost.…

jzmaddock:
…Log.

Fixes https://github.com/boostorg/multiprecision/issues/207.


cosurgi:
Try `using boost::multiprecision::log;` instead. It works for me :)

Issue #219, opened 2020-04-15
The floating_point_examples test requirement


eldiener:
For the floating_point_examples.cpp test you have the requirement:

`<toolset>gcc-mingw:<cxxflags>-Wa,-mbig-obj`

but gcc-mingw is not a toolset, although maybe it once was. I tried instead:

`<toolset>gcc/<target-os>windows:<cxxflags>-Wa,-mbig-obj`

but despite compiling with mingw-w64/gcc on Windows the -Wa,-mbig-obj was not being set, so I do not know what should work in that case. Of course if I add to my user-config.jam for gcc on Windows a <cxxflags>-Wa,-mbig-obj then it gets picked up from there, but I do not know how to get it picked up from your test requirement. However even when it does get picked up the mingw-w64/gcc-9.3 compiler still gives errors from the assembly. What can one expect on Windows <g>.

eldiener:
What do you mean by 'And it does the business for this application, so I’m not clear why it doesn’t work for you.' ?

If I run:

b2 toolset=gcc-9.3 etc.

it does not think that this is 'gcc-mingw', whatever that is supposed to mean, despite the fact that the gcc is on Windows and is a mingw-w64/gcc-9.3 implementation. It does in the gcc.jam file identify my gcc as mingw using its internal logic, and it does configure gcc with my user-config.jam statement of:

using gcc : 9.3 : 
    "my_path/my_command_etc" :
    <cxxflags>-Wno-unused-local-typedefs 
    <cxxflags>-ftrack-macro-expansion=0 
    <cxxflags>-Wno-unused-variable 
    <cxxflags>-D_GLIBCXX_USE_CXX11_ABI=1 
    <root>"c:/Utilities/mingw-w64/x86_64-9.3.0-posix-seh-rt_v7-rev0/mingw64"
    ;

correctly.

BTW, since you asked, I am building gcc-9.3 on Windows using MSYS2 from the implementation at https://github.com/niXman/mingw-builds . It has worked perfectly fine for me. I recomnmend it.

[...]

PR #222, opened 2020-04-18
Workaround for over-sized object files on Mingw.

jzmaddock:
Fixes https://github.com/boostorg/multiprecision/issues/219

### DIFF
--- a/config/Jamfile.v2
+++ b/config/Jamfile.v2
@@ -74,6 +74,7 @@ exe has_eigen : has_eigen.cpp ;
 exe has_f2c : has_f2c.cpp f2c ;
 obj has_is_constant_evaluated : has_is_constant_evaluated.cpp ;
 obj has_constexpr_limits : has_constexpr_limits_cmd.cpp : <cxxflags>-fconstexpr-ops-limit=268435456 ;
+obj has_big_obj : has_big_obj.cpp : <cxxflags>-Wa,-mbig-obj ;
 
 explicit has_gmp ;
 explicit has_mpfr ;
@@ -85,5 +86,6 @@ explicit has_mpc ;
 explicit has_eigen ;
 explicit has_is_constant_evaluated ;
 explicit has_constexpr_limits ;
+explicit has_big_obj ;
 explicit has_f2c ;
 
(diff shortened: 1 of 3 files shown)


jzmaddock:
@pabristow : you should see the command line as:

```
 "g++"   -fvisibility-inlines-hidden -m64 -mthreads -O0 -fno-inline -Wall -fvisibility=hidden -Wa,-mbig-obj -Wno-missing-braces -DBOOST_ALL_NO_LIB=1  -I"..\..\.." -I"..\include"  -c -o "..\..\..\bin.v2\libs\multiprecision\example\floating_point_examples.test\gcc-9.2\debug\debug-symbols-off\threading-multi\visibility-hidden\floating_point_examples.o" "floating_point_examples.cpp"
```

Which works for me with MSys64 GCC-9.2, but it is right on the limit :(

pabristow:
I can confirm that 9.2.0 works OK, so your fix is correct, for which thanks.

But GCC 10 looks like the straw that breaks the camels back :-(

Is it time to break down the floating_point_examples to smaller examples?

Shall I investigate?  Where does it report the fraction of table size used?

pabristow:
You asked about the command line, but it is now lost, and because I was on a detached HEAD, I seem not to have the log file for my failure.  Perhaps I can retry after you have merged big_obj changes to develop to make sure we are both using the same code?  What is the off option? that you added?

eldiener:
You can always see the command line which b2 generates to build anything by passing the -d2 switch to the b2 invocation.

I am waiting for this PR to merge before testing it out with gcc on Windows. I have all recent released versions of gcc on which I can easily test, but I do not have 10.0 since it has not been officially released. I can probably build 10.0 from https://github.com/niXman/mingw-builds but I have learned that for my sanity playing around with non-released software is rarely productive for me. But occasionally I do, as in Visual Studio 2019 Preview where Microsoft has a really easy to use GUI for keeping it up to date.

jzmaddock:
Merging now, like @eldiener I tend to ignore pre-release gcc versions as they often have multiple issues which ultimately get fixed.

eldiener:
I tested mingw-w64/gcc-9.3 on Windows 10 and your fix works for the floating_point_examples. Thanks !

PR #224, opened 2020-04-22
Improve special function support

jzmaddock:
And hook up testing to circle-ci as there are unnoticed regressions.


jzmaddock:
> I've been meaning to ask if we have a policy on the order of system include and Boost includes.

Not really, and it *shouldn't* make any difference - in this case I have a weird problem that's only showing up in C++17 and later that was fixed by re-ordering things.  At some point I'll try and investigate...

PR #225, opened 2020-04-23
Add .azure-pipelines.yml

jzmaddock:


jzmaddock:
This is hopeless: I had hoped that we could offload some of the Appveyor jobs to azure-pipelines and so speed up CI.  But this took an hour to build just 20 targets and then timed out, even just getting Boost installed took ages...

Closing down.

PR #231, opened 2020-05-03
Fast addition in cpp_int

madhur4127:
Compilers currently generate *bad* assembly so, with a little help to the compiler, I was able to reduce the number of instructions (`mov` and `add` both) in the addition loop. 

[See the Godbolt link showing the difference in assembly](https://godbolt.org/z/lH1_Hc)

Benefits:
- Can be easily unrolled
- Easy to read
- Efficient (?)

```
// clang 10.0, Arch Linux, -std=c++11
// Current develop
BM<bst_int>/40000        3826 ns         3730 ns       191511
BM<bst_int>/80000        4545 ns         4486 ns       157061
BM<bst_int>/120000       5416 ns         5335 ns       135902
BM<bst_int>/160000       6238 ns         6139 ns       113651
BM<bst_int>/200000       6924 ns         6829 ns       103996

// This PR
BM<bst_int>/40000        3395 ns         3362 ns       208408
BM<bst_int>/80000        3954 ns         3921 ns       172037
BM<bst_int>/120000       4551 ns         4511 ns       158986
BM<bst_int>/160000       5120 ns         5063 ns       136567
BM<bst_int>/200000       5545 ns         5496 ns       112399

BM<mpz_int>/40000         751 ns          744 ns       961070
BM<mpz_int>/80000        1536 ns         1522 ns       461803
BM<mpz_int>/120000       2389 ns         2371 ns       297695
BM<mpz_int>/160000       3475 ns         3439 ns       215895
BM<mpz_int>/200000       4008 ns         3972 ns       178096
```

Concerns:
- Most importantly will boost take this path of using intrinsics?
- Still unable to see huge performance gains as `mov` instructions are reduced and also `add` is executed only once instead of 4 times. Needs some investigation.
- GCC cannot unroll the loop while clang can, thus there is a huge disparity between GCC & LLVM.
- This assembly can be optimized further by writing assembly by hand (inline asm?) but still has potential.
- Portability [New!]

PR #233, opened 2020-05-05
Integrate PR for faster addition.

jzmaddock:
Splits unsigned_add up into multiple compiler/processor specific parts with selection logic to choose between them.
See https://github.com/boostorg/multiprecision/pull/231.


include/boost/multiprecision/cpp_int/addition/add_unsigned_addc_64.hpp
```diff
@@ -0,0 +1,89 @@
...
+         bool s = a.sign();
+         result = static_cast<double_limb_type>(*a.limbs()) + static_cast<double_limb_type>(*b.limbs());
+         result.sign(s);
+         return;
+      }
+      result.resize(x, x);
+      typename CppInt2::const_limb_pointer pa = a.limbs();
+      typename CppInt3::const_limb_pointer pb = b.limbs();
+      typename CppInt1::limb_pointer       pr = result.limbs();
+      //typename CppInt1::limb_pointer       pr_end = pr + m;
```
madhur4127, line 51: this can be removed

include/boost/multiprecision/cpp_int/addition/add_unsigned.hpp
```diff
@@ -0,0 +1,36 @@
...
+#ifdef __has_include
+#if __has_include(<immintrin.h>)
+#define BOOST_MP_HAS_IMMINTRIN_H
+#include <immintrin.h>
+#endif
+#endif
+
+#ifdef BOOST_MP_HAS_IMMINTRIN_H
+
+#if defined(__AVX__) && defined(BOOST_HAS_INT128)
```
madhur4127, line 20: As per intrinsics, I think this should be `__ADX__` ([link](https://clang.llvm.org/doxygen/adxintrin_8h_source.html))

Also please see this before using ADX extensions: [link](https://stackoverflow.com/a/45796549/819979)

jzmaddock:
Ugh, getting this correctly configured sucks :(

You're correct that gcc/clang don't support the adxc intrinsics, MSVC and Intel do so I've left the code in for them at present.... not convinced it makes much/any difference though.  It's a shame because I can see a use case for a double-addition-chain coming up.

Changing to adding 64-bit chunks in msvc makes only a slight difference: we appear to be right at the limit here, as a straight memcpy is only very slightly cheaper than the addition routine for me in msvc!

include/boost/multiprecision/cpp_int/addition/add_unsigned.hpp
jzmaddock: I've disabled AVX usage for gcc/clang since they don't generate AVX instructions anyway as you point out.  MSVC uses __AVX__ (see https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019), just need to check Intel/linux.

madhur4127: Intel IA32 Manual says to check CPUID bits for ADX extensions for ADCX, AODX.

It would be amusing if MSVC defined ability of ADX in Advanced vector extension (AVX, SIMD). SIMD won't work because there's a carry dependency between limbs. AFAIK there's no preprocessor macro for ADX in MSVC.

ADCX would be painful to made to work and would hurt performance in addition.

ADX was originally designed to support ultra fast naive multiplication of 512*512 bits.

jzmaddock: You're right! :(

madhur4127:
> Changing to adding 64-bit chunks in msvc makes only a slight difference: we appear to be right at the limit here, as a straight memcpy is only very slightly cheaper than the addition routine for me in msvc!

Then writing a generic function instead of 32/64 version would remove code duplication. I am thinking of writing switches for selecting the versions (based on limb_type) and cast type (based on the compiler).

jzmaddock:
> Then writing a generic function instead of 32/64 version would remove code duplication. I am thinking of writing switches for selecting the versions (based on limb_type) and cast type (based on the compiler).

Nod.  That's not a bad idea now that we're starting to get a handle on what compiler supports what.

jzmaddock:
Superseded by #234.

PR #240, opened 2020-05-25
rsqrt

NAThompson:


NAThompson:
Ok, strangely the MPFR `rsqrt` implementation is found in `namespace boost::math`, but the `eval_rsqrt` of `float128` is not. In any case, the following ulps plot show they both work:

```cpp
#include <boost/multiprecision/mpfr.hpp>
#include <boost/math/tools/ulps_plot.hpp>

int main()
{
    using boost::multiprecision::number;
    using PreciseReal =number<boost::multiprecision::mpfr_float_backend<400>>;
    using CoarseReal = boost::multiprecision::float128;
    using boost::math::tools::ulps_plot;
    std::string filename = "rsqrt_quad_03.svg";
    int samples = 10000;
    // How many pixels wide do you want your .svg?
    int width = 1400;
    // Near a root, we have unbounded relative error. So for functions with roots, we define an ULP clip:
    PreciseReal clip = 2.5;
    auto f = [](PreciseReal x) {
        using boost::math::rsqrt;
        return rsqrt(x);
    };
    auto plot03 = ulps_plot<decltype(f), PreciseReal, CoarseReal>(f, std::numeric_limits<CoarseReal>::min(), CoarseReal(3), samples);
    plot03.clip(clip).width(width);
    std::string title = "rsqrt ULPs plot at quad precision";
    plot03.title(title);
    plot03.vertical_lines(6);
    auto g = [](CoarseReal x) {
        CoarseReal xk = 1/std::sqrt(static_cast<long double>(x));
        // Single Newton iteration for f(x) = arg.value() - 1/x^2.
        return  xk + xk*(1-x*xk*xk)/2;
    };
    plot03.add_fn(g);

    // Two Newton iterates:
    auto g2 = [](CoarseReal x) {
        CoarseReal xk = 1/std::sqrt(static_cast<long double>(x));
        xk = xk + xk*(1-x*xk*xk)/2;
        return xk + xk*(1-x*xk*xk)/2;
    };
    plot03.add_fn(g2, "orange");
    plot03.write(filename);

    filename = "rsqrt_quad_3_100.svg";

    auto plot3_100 = ulps_plot<decltype(f), PreciseReal, CoarseReal>(f, CoarseReal(3), CoarseReal(100), samples);
    plot3_100.clip(clip).width(width);
    plot3_100.title(title);
    plot3_100.vertical_lines(6);
    plot3_100.add_fn(g);
    plot3_100.add_fn(g2, "orange");
    plot3_100.write(filename);
}
```

And here's the result we have on `[0,3]`:

Screen Shot 2020-05-25 at 4 44 41 PM

Screen Shot 2020-05-25 at 4 45 06 PM


If anyone has a second, could they run this on Windows to see if we need two iterations when `long double` is 64 bits?

NAThompson:
Ok, I'm a bit confused about where everything should be documented and tested. Normally this would be implemented in boost.math, and I'd put the ulps plot in the documentation there. But this is a kinda weird situation where it's half part of the `mpfr_wrapper`, half custom implementation.

Should I do another PR in boost.math? That would also get `rsqrt` into `float`, `double`, and `long double` precision.

jzmaddock:
>Ok, I'm a bit confused about where everything should be documented and tested. Normally this would be implemented in boost.math, and I'd put the ulps plot in the documentation there. But this is a kinda weird situation where it's half part of the mpfr_wrapper, half custom implementation.

>Should I do another PR in boost.math? That would also get rsqrt into float, double, and long double precision.

What I've done previously is provide the main/generic implementtion in boost/math/special_functions/ and then overload the function in multiprecision as required.

jzmaddock:
Ooops, forgot to say, this of course changes the namespace that rsqrt is declared in.

NAThompson:
@jzmaddock : Yup, just validated that this works and does indeed pick up the right version of the file. Tests in boost.math.

PR #244, opened 2020-06-05
Representations

NAThompson:


NAThompson:
This PR is more naturally part of the pslq PR.

PR #270, opened 2020-09-05
Improve the trig std lib functions

jzmaddock:
This addresses https://github.com/boostorg/multiprecision/issues/264 in large part.


NAThompson:
@jzmaddock : Wowsers you're going for half-ulp?

PR #276, opened 2020-12-12
reduce dec_float static init with constexpr constants

ckormanyos:
This draft PR investigates using selected features of C++11 in order to reduce the heavy load of static initialization associated with `cpp_dec_float`.

Historically `cpp_dec_float` has rquired certain static instances of class-local constants representing such things as min/max value of the class, epsilon, common small values such as 0.5, 1, 2, etc.

With some C++11 constructs (like `<initializer_list>`, `constexpr`) it might be possible to remove some or most of these static constant values. The run-time impact is expected to be performance-neutral. However, these changes reduce memory consumption (especially for higher digit counts) and significantly reduce the scope and complexity of the so-called static initialization that runs before the call to `main()`.

Aside: Future potential use improveements of `boost::math::constants` might possibly be able to be derived from this kind of technology, but these are not in the scope of this draft PR at the moment.

The checked in code handles nan, inf, max, eps.

### DIFF
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -16,9 +16,11 @@
 #ifndef BOOST_MP_CPP_DEC_FLOAT_BACKEND_HPP
 #define BOOST_MP_CPP_DEC_FLOAT_BACKEND_HPP
 
+#include <initializer_list>
+#include <limits>
+
 #include <boost/config.hpp>
 #include <boost/cstdint.hpp>
-#include <limits>
 #ifndef BOOST_NO_CXX11_HDR_ARRAY
 #include <array>
 #else
@@ -69,6 +71,11 @@ template <unsigned Digits10, class ExponentType, class Allocator>
 class cpp_dec_float
 {
  private:
+   static BOOST_CONSTEXPR_OR_CONST boost::uint32_t pow10_maker(boost::uint32_t n)
+   {
+     return ((n == UINT32_C(0)) ? UINT32_C(1) : pow10_maker(n - UINT32_C(1)) * UINT32_C(10));
+   }
+
    static const boost::int32_t cpp_dec_float_digits10_setting = Digits10;
 
    // We need at least 16-bits in the exponent type to do anything sensible:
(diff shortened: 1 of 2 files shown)


include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -69,6 +71,11 @@ template <unsigned Digits10, class ExponentType, class Allocator>
 class cpp_dec_float
 {
  private:
+   static BOOST_CONSTEXPR_OR_CONST boost::uint32_t pow10_maker(boost::uint32_t n)
```
jzmaddock, line 74: Since from the next release onwards we're moving to C++11 or later, this could just be straight `constepxr` and we can drop the macros now?

include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -284,25 +298,25 @@ class cpp_dec_float
    }
 
    // Specific special values.
-   static const cpp_dec_float& nan()
+   static BOOST_CONSTEXPR_OR_CONST cpp_dec_float nan()
    {
-      static const cpp_dec_float val(cpp_dec_float_NaN);
-      init.do_nothing();
-      return val;
+      return cpp_dec_float(cpp_dec_float_NaN);
```
jzmaddock, line 303: I wonder about this - I _think_ what you're really looking for here is `consteval`, but that's a C++20 feature :(  

I'm reasonably sure that the constructor here will be runtime evaluated (ie worst than before!) unless every *use* of the function is marked as `constexpr` which is error prone at best.  I think I'm right in saying that GCC will optimise this to a const-expression when used anyway, but it's not required to, and I'm fairly sure msvc does not do so.

So I think we either make these constants member values:

```
static constexpr cpp_dec_float nan = cpp_dec_float(cpp_dec_float_nan);
```

Which still requires an out of line definition (until C++17 anyway).  Or we do something like:

```
static constexpr const cpp_dec_float& nan()
{
  static constexpr cpp_dec_float val(cpp_dec_float_nan);
   return nan;
}
```

jzmaddock:
Hi Chris, this is a good idea IMO, but some of the details will be tricky to sort out - see my detailed comments.

ckormanyos:
Thank you for reviewing this, John. Your review points seem very clear, detailed and correct.

> I'm reasonably sure that the constructor here will be runtime evaluated (ie worst than before!)

I was also considering this. In particular, the default behavior of  `cpp_dec_float` uses `std::array` to contain its value field. So the performance hit of returning (on the fly, even with move-semantics) of  `std::array` will be more significant than I might have assumed. This run-time impact will be reduced if dynamic allocation is used.

Your intuition seems correct, as the overall performance will be either slightly worse or significantly worse.

ckormanyos:
> Which still requires an out of line definition (until C++17 anyway). Or we do something like:

```
static constexpr const cpp_dec_float& nan()
{
   static constexpr cpp_dec_float val(cpp_dec_float_nan);
   return val;
}
```

That is worth considering.

I am actually in the process of putting some of this stuff _on the metal_ within the context of some other research. I have C++11 and C++17 compilers from a variety of suppliers all competing to melt-down the RAM and ROM footprint of mega-digit calculations (another derivative work).

ckormanyos:
See summary in comments above.

Issue #278, opened 2020-12-20
Mixing different cpp_int_backend


mglisse:
The following program
```c++
#include <boost/multiprecision/cpp_int.hpp>

typedef boost::multiprecision::cpp_int T;
typedef boost::multiprecision::number<boost::multiprecision::cpp_int_backend<512>> U;

int main(){
  T a = T(1) << 150;
  T b = T(1) << 150;
  U c = a * b;
}
```
causes 3 allocations. 2 are normal, for a and b. However, when evaluating c, `a*b` is first evaluated in type T (with an allocation) and then copied to type U, instead of being directly computed in an object of type U. If I naively edit boost to make it call `do_assign(true_)`, and extend `setup_karatsuba` and `multiply_karatsuba` to allow mixed types, it seems to work as expected. It would be useful to be able to mix those types without paying such a penalty.

jzmaddock:
There's a couple of bugs, a big hole in our testing, and some documentation errors here... leave it with me.

jzmaddock:
See https://github.com/boostorg/multiprecision/pull/279.

mglisse:
Thanks, that seems to help a bit. I assume it is just the first step, I am still getting 3 allocations for the testcase. I guess another important step would be to relax `is_same` in `number::operator=(expression)` (and possibly adapt a few places?) so that it is possible to evaluate an expression directly into a variable of a slightly different type. But that may require some mechanism to advertise what types are compatible enough...

jzmaddock:
3 allocations in the test case is as low as it will ever get IMO: as there are 3 independent variables each with their own unique value (or am I missing something)?  The move occurs when the result of `a*b` is moved into `c`.  Previously this test case actually had 4 allocations.

Note you can also perform mixed arithmetic via:

```
multiply(c, a, b);
```

Which calculates `a*b` at the precision of `c` (and without a temporary).

mglisse:
> Previously this test case actually had 4 allocations.

Only 3, because U has a large enough minBits that it does not require an allocation. That's actually the whole point of using U here.

Yes, `U c;multiply(c,a,b);` works with your branch and has only 2 allocations. But the point of expression templates is to make natural code like `U c=a*b;` evaluate the same as whatever is best, probably `U c;multiply(c,a,b);` indeed in this case. I understand this may not be trivial, but it doesn't seem impossible either.

jzmaddock:
Ah yes I see.

[...]

PR #284, opened 2020-12-31
Allow mixed moves between related types.

jzmaddock:
Fix for issue https://github.com/boostorg/multiprecision/issues/278.

Allows more moves between related types, and direct operations on mixed types.

@mglisse : This reduces your example to 2 allocations, likewise similar operations.


mglisse:
Thanks, this looks great. Sorry it required so many changes in the comparison code...
Do you expect `is_compatible_mixed_arithmetic_type` may eventually become a documented interface for people writing their own backend? Or maybe you prefer to give it some time, since it isn't clear yet if this should take 2 backends or 2 numbers, or other variants.

jzmaddock:
>Do you expect is_compatible_mixed_arithmetic_type may eventually become a documented interface for people writing their own backend? Or maybe you prefer to give it some time, since it isn't clear yet if this should take 2 backends or 2 numbers, or other variants.

Probably, but more experimentation may be required first.

Right now I've just tried another sanity check - testing mixed arithmetic between `cpp_int` and `number<cpp_int_backend<N>>`  and there are gazillions of errors :(  The problem is that neither type is "better" than the other (ie they are both implicitly convertible to the other type since the precisions are the same) so the operator overloads are (rightly) all ambiguous.  I'm not sure how to fix this without yet another traits class which specifies the result type in case of ties like this, but it's not clear to me even what the default should be?

jzmaddock:
>Right now I've just tried another sanity check - testing mixed arithmetic between cpp_int and number<cpp_int_backend<N>> and there are gazillions of errors :( The problem is that neither type is "better" than the other (ie they are both implicitly convertible to the other type since the precisions are the same) so the operator overloads are (rightly) all ambiguous. I'm not sure how to fix this without yet another traits class which specifies the result type in case of ties like this, but it's not clear to me even what the default should be?

Ah, this is a non issue: if we've asked that two types be treated as equivalent then the notional result type of `a op b` is irrelevant and will be determined by the type we're assigning to anyway.  So we can just break the tie in the operator overloads either which way and everything should be fine.... testing now.

mglisse:
> Ah, this is a non issue: if we've asked that two types be treated as equivalent then the notional result type of `a op b` is irrelevant and will be determined by the type we're assigning to anyway. So we can just break the tie in the operator overloads either which way and everything should be fine....

Yes, as long as you use expression templates, you can use the type of the lhs. It is a bit strange for temporaries if the expression is complicated enough to require temporary variables, but not really a problem. And I'll propose a way to select a different type for temporaries later.
With a template parameter to disable expression templates, then you do have to choose a type... Not sure if we should arbitrarily pick the first one, or somehow combine them (max of the maxbits, etc).

jzmaddock:
@mglisse just updated the docs which completes this I think.  The added examples/benchmarks show some nice improvements too.

mglisse:
Yes, I was just reading the new doc, it looks great, thanks a lot.
I might have gone with `is_equivalent_number_backend<Backend1, Backend2>`, because it reduces the risk of instantiating `number<B>` before the partial specialization of `is_equivalent_number_type` is done (but I can't actually come up with a real issue this causes), and compatibility between the ET and non-ET versions of a type could be natural. I was also wondering about some trait like common_type, in case a mixed operation should return a 3rd type instead of just saying yes/no to compatibility, but that's a different feature. Those are just musings, I think the PR is fine as is (of course I didn't stress test it).
Once this is in, the next step I would be interested in is a way for a backend to advertise a different, equivalent backend that should be used for temporaries inside number: for instance the evaluation of `a*b<c*d` uses 2 temporaries, and I'd like to automatically use a different type for those.

PR #288, opened 2021-01-14
Huge PR that begins moving the library to C++11 and later.

jzmaddock:


ckormanyos:
This is amazing, John. Thank you for leading this incredible work.

It is great to see things like `<cstdint>`, `std::enable_if`, `static_assert` and `constexpr`. Regarding the word `constexpr` I believe, as the AppVeyor selection of tests indicates, that this lowly workd finally arrived on the MSVC scene for `_MSC_VER >= 1400`. Other features came sooned on this particular compiler.

jzmaddock:
Thanks Chris, just removed all MPL usage as well.... a bit of brain dead editing was just the ticket for the current lockdown! ;)  There might be a couple more changes to come, but this is mostly there now I think.

With regard to msvc: MSVC-14.0up3 first fully supported `constexpr` it seems, `noexcept` was also missing before that, and also vc12 had some issues with not finding the correct function overloads in some cases (enable_if not working properly), so yes, vc14 will be the first supported MSVC compiler after this change.

madhur4127:
Wow John, that's so much work. Thanks!

What's the C++ base standard that you plan to support? Like full comparability for C++11 for upcoming features or you want to move to new features like Boost.Math?

jzmaddock:
>What's the C++ base standard that you plan to support?

C++11.

We've signalled the change for a year now, so it's time to move on.  There are actually some things which would be both more efficient to compile and easier to implement with C++17's `if constexpr`, but that will have to wait for a few years yet!

In the mean time, something seems to have broken CI...

pabristow:
I'm not surprised the CI has indigestion!  Congratulations on a tremendous step forward and a major achievement in one giant leap.

I know we have given notice for over a year that it was going to happen, but now that it really will with the next release, I wonder if we need to give a much louder notice that the current release is their last release for people who are stuck with pre-c++11 compilers?

(and perhaps to give notice now that a move to minimum C++17 will take place two years hence to really focus peoples mind on the need to plan to 'KEEP UP'.)

jzmaddock:
>I know we have given notice for over a year that it was going to happen, but now that it really will with the next release, I wonder if we need to give a much louder notice that the current release is their last release for people who are stuck with pre-c++11 compilers?

Other than a very loud announcement in the release notes, I'm not sure what we can do?  I will add some #errors to the headers to catch C++03 compiles as well.

>(and perhaps to give notice now that a move to minimum C++17 will take place two years hence to really focus peoples mind on the need to plan to 'KEEP UP'.)

Might be a bit soon?  I don't think any of the main compilers are C++17 by default yet.... at the current rate it will be nearer 2030 before we can move again !!

ckormanyos:
>> I wonder if we need to give a much louder notice that the current release is their last release for people who are stuck with pre-c++11 compilers?

> Other than a very loud announcement in the release notes, I'm not sure what we can do? I will add some #errors to the headers to catch C++03 compiles as well.

I guess on a practical side, it might really be more clear to add some purposeful `#error` compile-time messages or else get potentially more confusing error messages when a non-C++11 compiler runs into either library headers from C++11 (like `<cstdint>`) or core language changes in C++11 such as `constexpr`.

NAThompson:
> Might be a bit soon? I don't think any of the main compilers are C++17 by default yet.... at the current rate it will be nearer 2030 before we can move again !!

They aren't default, but they are compatible. Default on the HPC systems is now gcc 9.3.

ckormanyos:
Here are some points we could still refine (and I'm happy to contribute to the editing of some, all or none of these as we may decide...)

1) Some constant expressions using `sizeof()` could be `constexpr`. An example is here: 
grafik

2) Some places could use either `constexpr` or sometimes in classes `static constexpr` for PODs depending on the context, such as here: 
grafik

3) A few places could make more refined use of `<type_traits>`. Not many, I found one here (`std::is_same<>::value` OK with `constexpr if`: 
grafik

4) We could change nearly 1,000 instances of `typedef` with `using`. This is in the spirit of modern C++11.

After that, i would like to take a more close look at move semantics and ensure that we are getting advantages full of these. I'm not so good at that language area, so need help here.

As mentioned, i would be happy to edit lots of this if desired. i am learning a lot on this issue.

jzmaddock:
>Some constant expressions using sizeof() could be constexpr. An example is here:

Nod. Lots of those, testing locally now...

>Some places could use either constexpr or sometimes in classes static constexpr for PODs depending on the context, 

Nod.   As above.

> A few places could make more refined use of <type_traits>. Not many, I found one here (std::is_same<>::value OK with constexpr if

I don't see that one: it's using `if constexpr` already in C++17 mode, and using `is_same` here would not be the same thing?

> We could change nearly 1,000 instances of typedef with using. This is in the spirit of modern C++11.

I tend to be a bit of a Luddite on things like this - it's a shiny new feature just the same as the old feature ;)  But it's probably just a search and replace away so I have no real strong feelings on the matter.

> After that, i would like to take a more close look at move semantics and ensure that we are getting advantages full of these. I'm not so good at that language area, so need help here.

Move semantics were reasonably well baked in from the start I hope, but I'm sure there's always more that could be done!

Many thanks for these.

ckormanyos:
>> A few places could make more refined use of <type_traits>. Not many, I found one here (std::is_same<>::value OK with `constexpr if`

> I don't see that one

You are right. My finding was wrong because `double` and `long double` are not `is_same`. My mistake.

ckormanyos:
>> We could change nearly 1,000 instances of typedef with using. This is in the spirit of modern C++11.

> on things like this - it's a shiny new feature just the same as the old feature ;) But it's probably just a search and replace away so I have no real strong feelings on the matter.

Unfortunately, this would be a more didicated edit effort because the order of `typedef` and `using` (an alias) is exactly reversed regarding which type is known and which is being defined. So we would have to fine each of 947 instances of `typedef`, switch to `using` and reverse the order of the types in the statements.

This is, in fact a shiny new thing that I would like to see and be proud to have. What if I handled this? Or ultimately decided to let it slide? Would you like the code with `using` or is it more like you resist and do not even want that new feature? I can live with either, but I`m willing to work for ùsing `using`...

I wonder if this is too technically challenging (or non-challenging?) for a GSoC 2021? The length of the GSoc has been cut in half this year to 6 weeks. This editing and testing effort could be the right thing... Or I could just do it myself, or let it go?

@jzmaddock: Thoughts on my ramblings...?

jzmaddock:
> Unfortunately, this would be a more didicated edit effort because the order of typedef and using (an alias) is exactly reversed regarding which type is known and which is being defined. So we would have to find each of 947 instances of typedef, switch to using and reverse the order of the types in the statements.

s/typedef\s+(.+)\s+([a-zA-Z0-9]+);/using $2 = $1;/g

Does the lot in about 15 seconds :)

ckormanyos:
>> a more didicated edit effort because the order of typedef and using (an alias) is exactly reversed

> s/typedef\s+(.+)\s+([a-zA-Z0-9]+);/using $2 = $1;/g

Awesome! Thank you for this.

jzmaddock:
OK I have one failure left that's a CI configuration (out of memory) issue that I'll fix shortly.

Other than that, I think this is ready to merge.  Anyone have any comments or last objections?

ckormanyos:
> this is ready to merge. Anyone have any comments or last objections?

It is great. Many thanks for the dedicated, detailed, eloquent migration. It's great to see so much beautiful C++11 in Multiprecision!

There is remaining work in CI for GHA (adding compilers and language specifications). These can be handled in #291

ckormanyos:
> There is remaining work in CI for GHA (adding compilers and language specifications).

Old GCC and clang compilers 5/6 have been added in #291. This brings the test matrices to the same level as those in Math. One test for C++1z on clang was disabled on source code level. After that the GHA are green.

The GCC/clang v5/6 tests on xenial in #291 can be merged into #288. Alternatively, we could mesynchronize these two branches and merge to develop, thereby I might suggest moving toward closing bot #288 and #291.

Thoughts?

PR #292, opened 2021-01-18
Prevent bootstrap fail from leaked CXX.

grafikrobot:


jzmaddock:
Thanks Rene: I've cancelled the travis/appveyor builds for this as they aren't effected by the fix, strangely I can find no trace a drone build for this, not even a failed one.

Please note that the drone failures were different from travis - bootstrap of b2 is passing, it's the b2 headers step that's failing.  There are also a lot of similar scripts out there :(

jzmaddock:
Got it!  Many thanks for tracking this down- it had me well stumped :(

PR #301, opened 2021-02-24
Fixes #289 with a clear intuitive example

### LINKED ISSUE #289: cpp_dec_float_100 can only be initialized precisely from a string? maybe should be documented, or what am I doing wrong?
Just started using Boost.Multiprecision and struggled to get a cpp_dec_float_100 initialized to some hardcoded value. Eventually got it working once I parsed a string containing that hardcoded value. The docs didn't mention this I just sort of got lucky and deduced it.

Should this be documented?
Is there a better way to do it lol?

```
#include <iostream>
#include "boost/multiprecision/cpp_dec_float.hpp"

int main(int argc, char *argv[])
{
    using namespace std;
    using EnormousFloat = cpp_dec_float_100;
    EnormousFloat i = 6942666.123456789098765432123456789098765432199999999999999999999999999999999999999999999999999999999999999999;
    cout << setprecision(numeric_limits<EnormousFloat>::max_digits10) << i << endl;
    //outputs: 6942666.12345678918063640594482421875

    string buffer("6942666.123456789098765432123456789098765432199999999999999999999999999999999999999999999999999999999999999999");
    istringstream is(buffer);
    EnormousFloat parsed;
    is >> parsed;
    cout << setprecision(numeric_limits<EnormousFloat>::max_digits10) << parsed << endl;
    //outputs: the actual full value as above/desired

    return 0;
}
```

ckormanyos:
Clear intuitive example has been added to the docs as discussed in the dialogue of #289.

### DIFF
--- a/doc/tutorial_float_builtin_ctor.qbk
+++ b/doc/tutorial_float_builtin_ctor.qbk
@@ -1,0 +1,59 @@
+[/
+  Copyright 2021 John Maddock.
+  Copyright 2021 Paul A. Bristow.
+  Copyright 2021 Christopher Kormanyos.
+
+  Distributed under the Boost Software License, Version 1.0.
+  (See accompanying file LICENSE_1_0.txt or copy at
+  http://www.boost.org/LICENSE_1_0.txt).
+]
+
+[section:floatbuiltinctor Construction from Built-In Floats]
+
+Construction of multiprecision types from built-in floating-point types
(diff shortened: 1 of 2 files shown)


ckormanyos:
I need some help later with generating all docs anew. I can build and do a syntax check, but I'm not sure when (and exactly how) to rebuild all docs and actually check them in.

PR #304, opened 2021-03-13
Robust GHA add repo

mborland:
Generally a duplicate of https://github.com/boostorg/math/pull/553 for multiprecision now that the GHA file is back.

### DIFF
--- a/.github/workflows/multiprecision.yml
+++ b/.github/workflows/multiprecision.yml
@@ -6,7 +6,14 @@
 # (See accompanying file LICENSE_1_0.txt or copy at http://boost.org/LICENSE_1_0.txt)
 
 name: multiprecision
-on: [ push, pull_request ]
+on:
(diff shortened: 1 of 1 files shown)


mborland:
@jzmaddock This is good to go. I have noticed that there has been a precipitous slowdown in GHA in the past week. Do you know if the boostorg account has worn out it's welcome (like Travis) now that most libraries have moved to it?

jzmaddock:
>Do you know if the boostorg account has worn out it's welcome (like Travis) now that most libraries have moved to it?

I have no idea, but I would expect there to be a limit to how many concurrent jobs boostorg can run at a time?

mborland:
I thought it used to be concurrent jobs per repo, but the [docs say](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) concurrent per account.

PR #307, opened 2021-03-17
Fix up ubuntu versions used on drone CI.

jzmaddock:


ckormanyos:
Thank you, John. It is really great and fascination to experience such great CI, and the technology strides ahead in 1.76.

Issue #313, opened 2021-03-30
UBSan warnings for cpp_dec_float


justusranvier:
The following test program will generate ubsan diagnostics if compiled with -fsanitize=undefined,integer using clang 11.1 and boost 1.75.0:

```c++
#include <boost/multiprecision/cpp_dec_float.hpp>

namespace mp = boost::multiprecision;

auto main() -> int
{
    using Float = mp::cpp_dec_float_100;
    auto test = mp::pow(Float{2}, 30) * mp::pow(Float{10}, -3);

    return 0;
}
```

```
/usr/include/boost/lexical_cast/detail/converter_lexical_streams.hpp:554:51: runtime error: unsigned integer overflow: 0 - 67108864 cannot be represented in type 'unsigned int'
    #0 0x258dad in bool boost::detail::lexical_ostream_limited_src<char, std::__1::char_traits<char> >::shr_signed<int>(int&) (/root/test/a.out+0x258dad)
    #1 0x258962 in boost::detail::lexical_ostream_limited_src<char, std::__1::char_traits<char> >::operator>>(int&) (/root/test/a.out+0x258962)
    #2 0x2585c7 in boost::detail::lexical_converter_impl<int, char const*>::try_convert(char const* const&, int&) (/root/test/a.out+0x2585c7)
    #3 0x25847a in bool boost::conversion::detail::try_lexical_convert<int, char const*>(char const* const&, int&) (/root/test/a.out+0x25847a)
    #4 0x25668a in int boost::lexical_cast<int, char const*>(char const* const&) (/root/test/a.out+0x25668a)
    #5 0x253de3 in boost::multiprecision::backends::cpp_dec_float<100u, int, void>::rd_string(char const*) (/root/test/a.out+0x253de3)
    #6 0x253b8f in boost::multiprecision::backends::cpp_dec_float<100u, int, void>::operator=(char const*) (/root/test/a.out+0x253b8f)
    #7 0x25219e in boost::multiprecision::backends::cpp_dec_float<100u, int, void>::cpp_dec_float(char const*) (/root/test/a.out+0x25219e)
    #8 0x24f5e4 in boost::multiprecision::backends::cpp_dec_float<100u, int, void>::min() (/root/test/a.out+0x24f5e4)
    #9 0x24b89d in boost::multiprecision::backends::cpp_dec_float<100u, int, void>::initializer::initializer() (/root/test/a.out+0x24b89d)
[...]

PR #314, opened 2021-03-31
Try first draft singed/unsigned exp/mod repair

ckormanyos:
Try using `ptrdiff_t` instead of `size_t`. TBD: If CI OK, then will also add a specific test case


ckormanyos:
Hi @jzmaddock as we progress on this, could we add some `-fsanitize=undefined` runs on either GCC, clang or a few combinations of  both?

I can't find out exactly how/if you put any sanitizers in circle or drone. I thought you had, but I'm not recalling where or how...

ckormanyos:
Hi @jzmaddock in CI running at the moment is a somewhat new idea in `cpp_dec_float`, which was closely related to this issue enough for me to try out.

Long have I noted the plethora of string constructors in the function `pow2`. I have added a new constructor from `initializer_list` and shown its potential use in a new form of [new form for pow2 here](https://github.com/boostorg/multiprecision/blob/e1fbbfaebf00ab1e413343861dd657a2633c0bc8/include/boost/multiprecision/cpp_dec_float.hpp#L2371).

At the moment, I am only using 9 table entries instead of the familiar 256. I wonder what you think, in particular of the ctor from `initializer_list`? In the future, this ctor could also facilitate `constexpr` construction as soon as the underlying data containers are adapted accordingly.

ckormanyos:
It appears as though the failing tests in the CI builds/runs are exclusively on OSX 7.x, which is seemingly not aware of the attribute `thread_local`. I think this needs to be changed in Boost.Math.

Something like in `config.hpp`, to ensure that `thread_local` is not attempted to be used for `(__clang__major__ <= 7)`, or similar such check.

jzmaddock:
>At the moment, I am only using 9 table entries instead of the familiar 256. I wonder what you think, in particular of the ctor from initializer_list? In the future, this ctor could also facilitate constexpr construction as soon as the underlying data containers are adapted accordingly.

Looks good to me!

jzmaddock:
>It appears as though the failing tests in the CI builds/runs are exclusively on OSX 7.x, which is seemingly not aware of the attribute thread_local. I think this needs to be changed in Boost.Math.

Ah... I think this was a deliberate change to drop support for non-C++-11 compilers, XCode-7 is pretty old now, should we care about this does everyone use the latest XCode release?

ckormanyos:
> a deliberate change to drop support for non-C++-11 compilers, XCode-7 is pretty old now

On the one hand, it looked like failures only for the attribute `thread_local`, which that compiler did not support. On the other hand, I belive we're talking about that 5,6 years old timing now. I personally can live without suppoting it in CI, or at all for that matter. In fact, I almost edited the drone scripts to get rid of it ini the branch just to get my own stuff green.

My vote would be to drop it, if that is too radical, we would need to see how to blend out `thread_local` and still get green on it, which might be simple or difficult, I can't judge that in the moment. But it looked like those were the only failing tests, the ones involving that attribute.

NAThompson:
> Ah... I think this was a deliberate change to drop support for non-C++-11 compilers, XCode-7 is pretty old now, should we care about this does everyone use the latest XCode release?

Yeah everyone uses the latest. I'd get rid of it.

ckormanyos:
This PR is now ready for review.

The fix addresses the original issue, by:
- changing the [signedness of the exp/mod](https://github.com/boostorg/multiprecision/blob/26f704dd1ec554d3a7dfd9b286cf13fc22502c85/include/boost/multiprecision/cpp_dec_float.hpp#L2019) calculation
- adding a [private constructor](https://github.com/boostorg/multiprecision/blob/26f704dd1ec554d3a7dfd9b286cf13fc22502c85/include/boost/multiprecision/cpp_dec_float.hpp#L559) from `std::initializer_list` which [subsequently simplifies `pow2`](https://github.com/boostorg/multiprecision/blob/26f704dd1ec554d3a7dfd9b286cf13fc22502c85/include/boost/multiprecision/cpp_dec_float.hpp#L2378).
- adding [test code exercising the original issue](https://github.com/boostorg/multiprecision/blob/ubsan_size_t_exp_issue313/test/git_issue_313.cpp)
- adding some [`-fsanitize=undefined`](https://github.com/boostorg/multiprecision/blob/26f704dd1ec554d3a7dfd9b286cf13fc22502c85/.github/workflows/multiprecision.yml#L142) runs in CI on GHA
- fixing CI run names

The scope of change was a bit larger than initially anticipated, but seems to go in the direction of fixing this issue and also simplifying the ovreall `cpp_dec-float.hpp`.

jzmaddock:
Looks good to me Chris - do you want to go ahead and merge when you're ready?

BTW the original issue was triggered by -fsanitize=integer.  However, adding that would trigger countless errors in our dependencies :(

ckormanyos:
> go ahead and merge when you're ready?

Yes. Done. Thanks for checking this.

ckormanyos:
Fixes #313 (in the Multiprecision part)

PR #316, opened 2021-04-04
Remove unused dependencies

mborland:
Boost.assert and Boost.throw_exception are both included but unused in `number.hpp`


mborland:
@jzmaddock The CI on this one is clean.

PR #323, opened 2021-05-10
Cleanup more cpp dec float

ckormanyos:
This is a partial rework of #315.

The new constructor of `cpp_dec_float` from `std::initializer_list` can cause all kinds of ambiguity problems. These include confusion with C++11 uniform initialization syntax and ambiguity with `cpp_dec_float`'s constructor from `std::string` (which also has construction from initializer list).

In light of this, replace the private new constructor from initializer list with a (still not perfect) private convenience function called `from_lst`. Use this consistently within the internal `cpp_dec_float` code.


ckormanyos:
Although potential constructor ambiguity problems were, in fact, shielded by the fronend `number` template, it is probably best to avoid constructors of the backend type from as many aggregate type(s) as possible.

Issue #325, opened 2021-05-19
cpp_dec_float broken on develop


jzmaddock:
@ckormanyos : cpp_dec_float appears to be limited to 17-digit precision on develop, test case is:

```
   using dec35 = boost::multiprecision::number<boost::multiprecision::cpp_dec_float<35>>;

   dec35 a = tgamma(dec35(2.5));
   dec35 b("1.3293403881791370204736256125058588870981620920917903461603558423");

   std::cout << fabs((a - b) / b) << std::endl;
```

The problem appears to have been introduced in https://github.com/boostorg/multiprecision/commit/aa88310934e3f7555c8a630269a18f555af88ac0 but I don't see why/where.

You can also reproduce with:

```
cd libs/multiprecision/test
../../../b2 specfun_cpp_dec_float
```

Which shows up lots of related failures (and I must add that to one of the CI runs, we haven't done so before because it takes ages to run).

ckormanyos:
> `cpp_dec_float` appears to be limited to 17-digit precision on develop,

Oh great catch John. My bad.

I think I simply got too hasty in removing `#include <boost/multiprecision/detail/big_lanczos.hpp>`.

I am not sure if that is the _only_ thing that I might have unintentionally broken.

I will correct in the branch then get those tests cycling.

ckormanyos:
I checked large digit range specifically in the multiprecision tests. But I wonder why this case slips through multiprecision's testing?

I will also soon try to add explicit tests in Multiprecision for low-digit count tgamma with dec float.

ckormanyos:
This is being handled in #326

ckormanyos:
The original issue had been the non-inclusion of the _big Lanczos_ header file, formerly known from Multiprecision.

As it turned out, the re-inclusion of this header file, did, in fact, lead to recognition of some imprecise Lanczos tables in selected digit ranges such as 34 decimal digits. In this way, a related weakness was found via this unrelated bug.

The Lanczos coefficients in most digit ranges have now been reworked in #326. Additional testing and verification carried out in #327.

Will now close this issue.

@jzmaddock Thanks again for such widespread work and excellent repairs on this issue and the scope of possible improvements to which it has brought us.

PR #327, opened 2021-05-21
Improve CI testing of tgamma at various precisions.

jzmaddock:
Also run the special function tests for select compilers.
NOTE: this still has some failures which need investigating.


ckormanyos:
Hi @jzmaddock what I do see in test failures are sporadic over-steppings of tolerance(s) for bin float at 80 or 100 decimal digits.

These are probably beyond or at the very top limit of Lanczos approximation. Recalling that dec float has quite a few more guard digits and that around that 100 digit region, we are either in Lanczos upper reaches or switching over to asymptotic (lossy) Stirling's approximation, I wonder if simply the tolerance needs to be raised for bin float?

I was unable to extract an exact test case for microscopic inspection in a standalone test case on my work station. Do you think it would be possible to isolate an exact failing test case regarding the exact arguments and tolerance being used?

jzmaddock:
Nod.  Investigating now.

The code that generated the approximations had an upper limit on the test data used to train the approximations, so they're really good within the test range of the generator, and then fade.  That's fine for double etc due the limited range of those types, but once you have int-sized exponents, things get more extreme.  I have a better approximation generated at 80-bit, but it fades for lgamma but not tgamma for reasons I don't understand yet.  And yes, switching to the asymptotic approximation is probably the ultimate fix, albeit I need to write some more tests first.  I'd like to try and understand things a bit more first, before I just stick an asymptotic sticking plaster over things too ;)

ckormanyos:
>  I'd like to try and understand things a bit more first, before I just stick an asymptotic sticking plaster over things too

Indeed. That is really strong --- and the better way to go if possible, as those Lanczos approximations are super efficient.

If I can help, please contact me. I also have a program that generates Lanczos coefficients, but for some reason I always lose something like half the precision of the type when I use the coefficients that are generated from my program. I have never really found the right mojo for the _N_ and _g_ parameters.

jzmaddock:
> I also have a program that generates Lanczos coefficients, but for some reason I always lose something like half the precision of the type when I use the coefficients that are generated from my program. I have never really found the right mojo for the N and g parameters.

Nod.  Lanczos is notorious for being ill-conditioned.   However, there are reasonably well known "sweet spots" for N and g (and you can find more for larger N) which you can scan through for whichever one gives the best-conditioned result at fixed precision.  That's what this program does: https://github.com/boostorg/math/blob/develop/tools/lanczos_generator.cpp.  Note that while it prints out the "best" approximation at the end, it's better to use some subjective judgement as there may be one with nearly the same error, but many fewer terms.

And finally.... the issue I had with lgamma and my revised 80-digit lanczos approximation was this: we use a rewrite of the Lanczos approximation in terms of z-1 or z-2 near 1 and 2 respectively, with differently generated coefficients for those special cases.  It turns out - and this should have been obvious with hindsight - that the best general approximation, isn't necessarily the best for the rearranged form near 1 and 2.  So I need to adjust the generator program to print the error rates in those special cases, and then I can hopefully just cut and paste the best bits.

ckormanyos:
>  issue I had with lgamma and my revised 80-digit lanczos approximation was this: we use a rewrite of the Lanczos approximation in terms of z-1 or z-2 near 1 and 2 respectively, with differently generated coefficients for those special cases. It turns out - and this should have been obvious with hindsight - that the best general approximation, isn't necessarily the best for the rearranged form near 1 and 2. So I need to adjust the generator program to print the error rates in those special cases, and then I can hopefully just cut and paste the best bits.

What an awesome analysis! Right on.

I don't think this really makes sense for 80, 100, or more decimal digits, but sometimes for smaller digit ranges on timy platforms (as you know I'm often hammering with math stuff) at those 1, 2 points on the real axis, I look into stuff like `N[Series[1/Gamma[x], {x, 1, 12}], 21]`. It's cool for float, but not so cool for 80 digits.

ckormanyos:
Hi @jzmaddock Sorry if I misunderstood. Some test failures seem to be in `cpp_dec_float-hpp`, which still (and maybe this was a misunderstanding) has `#include <boost/multiprecision/detail/big_lanczos.hpp>`.

jzmaddock:
> Sorry if I misunderstood. Some test failures seem to be in cpp_dec_float-hpp, which still (and maybe this was a misunderstanding) has #include <boost/multiprecision/detail/big_lanczos.hpp>.

Oh :(   It was passing locally, but changes to develop was putting the #include back in - fixed now.

jzmaddock:
The one failure looks spurious.... merging.

ckormanyos:
This is great, John. Thank you. I really appreciated your program to calculate Lanczos coefficients and learned a lot as well.

It's always amazing what improvements we find in the wiley... `tgamma` over a wide range of precision and argument.

ckormanyos:
Indeed. These improvements also fit well with potential forward motion on double-double, quad-double and the like, as these are in the Lanczos precision range.

ckormanyos:
This PR has fixed #325

Issue #324, opened 2021-05-16
Enum class conversion broken since 1.76.0


wak-google:
Consider the following code snippet (`g++ -std=c++17 -o main main.cpp`)
```
#include <boost/multiprecision/number.hpp>
#include <boost/multiprecision/cpp_int.hpp>

template <unsigned N>
using fixed_uint_t =
    boost::multiprecision::number<boost::multiprecision::cpp_int_backend<
        N, N, boost::multiprecision::unsigned_magnitude,
        boost::multiprecision::unchecked, void>>;

enum class Status : uint8_t
{
    ok = 0,
    unknown = 1,
};

int main()
{
    auto b = static_cast<fixed_uint_t<7>>(Status::ok);
    return 0;
}
```

This used to compile with 1.75.0, but now produces
```
main.cpp: In function ‘int main()’:
main.cpp:18:53: error: no matching function for call to ‘boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<7, 7, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>, boost::multiprecision::et_off>::number(Status)’
   18 |     auto b = static_cast<fixed_uint_t<7>>(Status::ok);
      |                                                     ^
In file included from main.cpp:1:
include/boost/multiprecision/number.hpp:317:50: note: candidate: ‘template<class Other, boost::multiprecision::expression_template_option ET> constexpr boost::multiprecision::number<Backend, ExpressionTemplates>::number(boost::multiprecision::number<Other, ET>&&, typename std::enable_if<(std::is_convertible<Other, Backend>::value && (! boost::multiprecision::detail::is_restricted_conversion<Other, Backend>::value))>::type*) [with Other = Other; boost::multiprecision::expression_template_option ET = ET; Backend = boost::multiprecision::backends::cpp_int_backend<7, 7, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>; boost::multiprecision::expression_template_option ExpressionTemplates = boost::multiprecision::et_off]’
  317 |    BOOST_MP_FORCEINLINE BOOST_MP_CXX14_CONSTEXPR number(number<Other, ET>&& val,
      |                                                  ^~~~~~
include/boost/multiprecision/number.hpp:317:50: note:   template argument deduction/substitution failed:
main.cpp:18:53: note:   mismatched types ‘boost::multiprecision::number<Backend, ExpressionTemplates>’ and ‘Status’
[...]

PR #333, opened 2021-06-04
Fix up conversions to and from enum's and add new tests.

jzmaddock:
Fixes: https://github.com/boostorg/multiprecision/issues/324.


ckormanyos:
Thank you John. Nice!

PR #334, opened 2021-06-06
Cleanup hash_combine usage.

jzmaddock:


include/boost/multiprecision/detail/hash.hpp
```diff
@@ -11,23 +11,23 @@
...
+}
+
 inline void hash_combine(std::size_t&) {}
 
 template <typename T, typename... Args>
 inline void hash_combine(std::size_t& seed, const T& v, Args... args) 
 {
     // gmp types require explicit casting
-    seed = static_cast<std::size_t>(static_cast<T>(seed) ^ (v + 0x9e3779b9 + (seed<<6) + (seed>>2)));
+    seed = seed ^ (hash_value(v) + 0x9e3779b9 + (seed<<6) + (seed>>2));
```
ckormanyos, line 27: Hi John. I have experienced difficulties with this line, but never discussed them. For me it is difficult to intuitively understand the operator precedence and the types of the integers. Are they meant to be int32_t or int? These are different things. And sure there are clear precedence rules, but xor, add, shift, it's confusing. I tried to add parentheses to this line in offline experiments once and completely failed. Would it be possible to clarify operator precedence, integral types with type information and parantheses?

jzmaddock: Nod.  The original code from Boost.Hash uses std::size_t throughout.  When I changed things to make the function vararg I broke that, which resulted in a slew of compiler warnings :(

This patch makes everything size_t's again since the `0x9e3779b9` is implicitly unsigned.  There shouldn't be any operator precedence issues now since `(seed<<6)` is still a size_t.  So we have the addition of 3 size_t's and an unsigned in `(hash_value(v) + 0x9e3779b9 + (seed<<6) + (seed>>2))` which is then xor'ed with `seed`.

jzmaddock: But I see I've broken __int128 support since std::hash doesn't support that.

jzmaddock:
Failures are unrelated to this PR, so merging.

Issue #97, opened 2018-11-15
Multiprecision Complex: to_string() overload?


NAThompson:
Would implementing `to_string()` for complex multiprecision numbers be a nice roundout of the feature set?

jzmaddock:
There's a str() member function already, or did you mean implement a version of std::to_string?  The latter seems to have rather unexpected behaviour?

NAThompson:
I meant to implement a version of `std::to_string` using ADL, something like `std::string to_string(complex128 z) { return z.str(); }`.

Then people would be able to write `std::string msg = "Error: z = " + to_string(z);`.

Would this be useful/worth the effort?

jzmaddock:
Nod.  On reflection it does help with generic code.  You would need to play with precision and format flags a little to match the behaviour of the std version, but in principle would it just be the same as:

```
template <class Backend, expression_template_option ExpressionTemplates>
std::string to_string(const number<Backend, ExpressionTemlates>& val)
{
  return val.str(6, std::ios_base::fixed|std::ios_base::showpoint);
}
```

?

NAThompson:
Just looked up the definition of `to_string()` [here](http://www.cplusplus.com/reference/string/to_string/):

> As many digits are written as needed to represent the integral part, followed by the decimal-point character and six decimal digits.
inf (or infinity) is used to represent infinity.
nan (followed by an optional sequence of characters) to represent NaNs (Not-a-Number).
The representations of negative values are preceded with a minus sign (-).

So it looks like your code is correct, though you could argue that this is a pretty bizarre spec (which I hadn't realized before opening this issue).

agurtovoy:
The cplusplus.com definition is somewhat incorrect; for all the numeric `to_string` overloads, the standard simply says (24.3.4.7 [string.conversions]):

*Returns*: Each function returns a string object holding the character representation of the value of its argument that would be generated by calling `sprintf(buf, fmt, val)` with a format specifier of `"%d"`, `"%u"`, `"%ld"`, `"%lu"`, `"%lld"`, `"%llu"`, `"%f"`, `"%f"`, or `"%Lf"`, respectively, where `buf` designates an internal character buffer of sufficient size.

My guess is that the "followed by the decimal-point character and six decimal digits" part of the cplusplus.com definition comes from [`sprintf`](https://en.cppreference.com/w/cpp/io/c/fprintf) `"%f"` spec, which sets the default precision for _floats_ to 6 (which makes sense for floats).

[...]

PR #339, opened 2021-06-21
Implement to_string via ADL.

NAThompson:
Solves #97


NAThompson:
@jzmaddock : I admit to not knowing how this should be tested, but otherwise it seems fine. Build error is a timeout.

jzmaddock:
The examples here might make some minimal tests: https://en.cppreference.com/w/cpp/string/basic_string/to_string though the 1e40 example will obviously change it's output depending on the precision of the type.

Can you add some minimal docs to doc/reference_number.qbk ?

My only (slight) concern, is that floating point values with large (64-bit!!) exponent ranges could generate stupidly long strings - something that might warrant a warning in the docs.

PR #346, opened 2021-07-18
Add some tests for long double and float128

cosurgi:
I am a bit tired today, so I will push what I have done so far with this. The `long double` appears to work, while `float128` mostly has some problems with the constructors.


cosurgi:
Oops, somehow by mistake I opened PR to the wrong target repo.

PR #348, opened 2021-07-25
Make float128 quiet_NaN constexpr

mborland:
Constructing from a `const char*` is not constexpr; replace with `NAN`.


mborland:
The only CI failure is from a killed build.

PR #350, opened 2021-07-28
Update gitignore

mborland:
Will reduce cmake and build related noise


mborland:
I am not entirely sure why the CI did not run, but I don't think there is a need to cycle the CI on this one.

jzmaddock:
Very strange failures!  Merging anyhow...

PR #353, opened 2021-08-04
Fix rational istream operator for platforms with unsigned char.

awulkiew:
```
../boost/multiprecision/number.hpp:2293:16: warning: result of comparison of constant -1 with expression of type 'char' is always true [-Wtautological-constant-out-of-range-compare]
../boost/multiprecision/number.hpp:2309:19: warning: result of comparison of constant -1 with expression of type 'char' is always true [-Wtautological-constant-out-of-range-compare]
```
If you think that this should be fixed in a different way feel free to close this PR.


jzmaddock:
Looks good to me, thanks for that!

Issue #368, opened 2021-09-18
Wrong division result with cpp_dec_float


dlaugt:
I'm using cpp_dec_float to don't deal on rounding issues when mathematic operations give some results with an exact base-10 representation. The result of 69000 / 184 = 375. With cpp_dec_float, it is a bit more.

```cpp
#include <iostream>
#include <boost/multiprecision/cpp_dec_float.hpp>

using decimal = boost::multiprecision::number<boost::multiprecision::cpp_dec_float<10>,
                                              boost::multiprecision::et_off>;

int main (int argc, char* argv[])
{
  decimal a = decimal {"69000"} / decimal {"184"};
  decimal b = ceil (a);

  std::cout << "a = " << a << std::endl;
  std::cout << "b = " << b << std::endl;
}
```

The above program shows:

> a = 375
> b = 376

ckormanyos:
> The above program shows:

```
    a = 375
    b = 376
```

Hmmm... For some reason, I can not, at the moment reproduce this behavior. I tried with Boost 1.77 and both VC14.2 and GCC 9.3 on x86_64-linux-gnu and got `375` for both `a` as well as `b`.

Could you please provide more information on the compiler/Boost-version you are using to obtain this result?

jzmaddock:
I'm fairly sure this is a result of the cpp_dec_float not performing any rounding (and having guard digits), but @ckormanyos will no doubt confirm shortly.

ckormanyos:
> `cpp_dec_float` not performing any rounding

Thanks john. That's what I thought at first, ... but...

> but @ckormanyos will no doubt confirm shortly

... actually, I am having difficulty reproducing the result. I obtain `375` for both `a` and `b`. I tried the compilers/Boost-version mentioned above and also develop branch of Boost.

So I would like to find out the compiler/Boost-version/target-system exhibiting this phenomenon.

dlaugt:
I'm using boost 1.73 with visual studio 16.11.2. The compilation is done in 32 bits.

I've tried with more decimals (50). This time, the result is smaller than 175.

```cpp
#include <iostream>
#include <boost/multiprecision/cpp_dec_float.hpp>

using decimal = boost::multiprecision::number<boost::multiprecision::cpp_dec_float<50>,
                                              boost::multiprecision::et_off>;

int main (int argc, char* argv[])
{
  decimal a = decimal {"69000"} / decimal {"184"};
  decimal b = floor (a);
  decimal c = ceil (a);

[...]

PR #371, opened 2021-09-21
Try first cut at issue 368

ckormanyos:
This PR attempts to address #368 with minimally non-intrusive changes of rounding in selected (but not all) places.


ckormanyos:
Hi @jzmaddock what's your opinion on this cut?

I introduced rounding-up in selected areas only including `floor`/`ceil`/cast-to-built-in-integral.

Advantages include:
- Handle the matter of #368 and possibly a subclass of related issues.
- Introduce some rounding into `cpp_dec_float` to be expanded later.
- Gain experience.

Disadvantage is that rounding has not yet been successfully applied to arithmetic operations and there are some open questions regarding the best way to do this. This could/should be completed within the framework of also refining equality on the sub-limb level.

At the moment, however, there are no tangible disadvantages other than items that were already missing anyway. And CI is passing (except Circle CI has been spinning around for a day or so).

So what do you think of this PR, its scope of change and if we should actually merge?

 is

test/git_issue_368.cpp
```diff
@@ -0,0 +1,63 @@
...
+      {
+        const std::uint32_t lo_hi = lo_index * hi_index;
+
+        const decimal_type a = decimal_type { lo_hi } / decimal_type { lo_index };
+        const decimal_type b = ceil(a);
+
+        const std::string str_a = boost::lexical_cast<std::string>(a);
+        const std::string str_b = boost::lexical_cast<std::string>(b);
+
+        result_is_ok &= (str_a == str_b);
```
jzmaddock, line 31: Questions: why are we comparing strings rather than values, and should we also test inputs which are not integer?  Negative values as well, plus floor too I suspect ;)

include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -3091,37 +3236,73 @@ template <unsigned Digits10, class ExponentType, class Allocator>
...
+   else
+   {
+      result.eval_round_self();
 
-   if (x.isneg())
-      result -= cpp_dec_float<Digits10, ExponentType, Allocator>::one();
-   result = result.extract_integer_part();
+      if(result.isint())
+      {
+         ;
```
jzmaddock, line 3251: ?

include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -522,6 +522,141 @@ class cpp_dec_float
...
+      if(needs_rounding)
+      {
+         using local_size_type = typename array_type::size_type;
+         using local_limb_type = typename array_type::value_type;
+
+         std::int32_t digits_limb_0 = 0;
+
+         local_limb_type tmp_limb_0 = data[0U];
+
+         // Manually count the number of base-10 digits on the zero'th limb.
```
jzmaddock, line 538: This is the bit that caused me to assume that correct round was next to impossible to do efficiently in this type, but maybe it's not so bad?  I can't help but wonder if there are more efficient implementations - maybe binary search comparisons?  Or can we even do table lookup based on the most-significant-bit?  Just thinking out loud here....

test/git_issue_368.cpp
ckormanyos: > why are we comparing strings rather than values, 

In this PR, there is rounding on `floor`, `ceil` and cast-to-int only. I removed the use of strings. But the comparison with, say, the result of `floor` or `ceil` is only true if the value being compared is casted to int prior to comparison. This is now done without strings in the candidate code of the PR's test file.

> we also test inputs which are not integer

It seems like there are already-existing rounding tests which do this.

> Negative values as well, plus floor too I suspect

Understood. Done in resent push(es).

include/boost/multiprecision/cpp_dec_float.hpp
ckormanyos: > `if(result.isint())`
> ?

Indeed. These sequences have been simlified in recent push(es).

include/boost/multiprecision/cpp_dec_float.hpp
ckormanyos: > the bit that caused me to assume that correct round was next to impossible to do efficiently in this type

Well, it sure is proving to be unruly. But there are some new efficiency provisions.
- There was an existing binary search which had calculated the number of base-10 digits in the leading limb. This is used instead of a manual counting loop now.
- For the power-of-10, a special non-`constexpr` table lookup has also been created for quick calculations of the power(s) of ten in similar matters.

ckormanyos:
Hi @jzmaddock I have attempted to address the review questions one-by-one.

Personally, I am skeptical to introduce _partial_ rounding on `floor`, `ceil`, and cast-to-int only.

The questions I ask myself are:

1. Is `cpp_dec_float` worse off with partial rounding?
2. Is `cpp_dec_float` neutral or about the same?
3. Or does this particl rounding improve the overall class without high risk of breaking anything else?

I tend to think we are around 2 and 3. I would, however, treat this skeptically. We deal with #368. But it seems like a partial treatment of rounding, and partial could be interpreted as wrong or less right than uniform truncation.

I ran some tests with rounding on arithmetic operatory, but found that this leads to quite a few changes in tolerances and behaviors that are already possibly running. I am not sure if rounding over all (also on arithmetic operations) would break existing code or not? I also came to believe that rounding only makes sense if we combine it with a more exact reatment of equality (equality on the sub-limb, digit level). This has also been avoided over the years.

I have no problem keeping these changes in the PR or letting it sit for a while, or even dropping this effort. I am, however, somewhat motivated by the urge to deal with #368.

Thoughts?

Issue #364, opened 2021-08-31
Standalone?


NAThompson:
Heads up: Just had a use case for boost/multiprecision as a standalone.

ckormanyos:
> boost/multiprecision as a standalone

That is in my opinion a good idea. I'm not very good at that kind of thing, but I'd help try for it.

Some of my other projecs use parts of Math and Multiprecision. Prior to C++11 and standalone Math, I needed to get over a hundred submodules, not it's much lower. Standalone Multiprecision would really help me in CI of several of my unrelated-to-Boost projects.

jzmaddock:
I think we're getting closer - I have a better rational_adaptor in the works which will remove the dependency to boost.rational.

mborland:
@NAThompson is it a safe assumption that standalone Multiprecision can depend on standalone Math? Here is the current breakdown of dependencies:

```
Primary dependencies for multiprecision:

assert:
    <boost/assert.hpp>
        from <boost/multiprecision/cpp_int/cpp_int_config.hpp>

config:
    <boost/config.hpp>
        from <boost/multiprecision/cpp_dec_float.hpp>
        from <boost/multiprecision/detail/atomic.hpp>
        from <boost/multiprecision/detail/constexpr.hpp>
        from <boost/multiprecision/float128.hpp>
        from <boost/multiprecision/number.hpp>
        from <boost/multiprecision/traits/explicit_conversion.hpp>

core:
    <boost/core/empty_value.hpp>
        from <boost/multiprecision/cpp_int.hpp>
    <boost/core/no_exceptions_support.hpp>
        from <boost/multiprecision/detail/default_ops.hpp>
        from <boost/multiprecision/detail/functions/pow.hpp>
    <boost/core/nvp.hpp>
        from <boost/multiprecision/detail/number_base.hpp>

integer:
    <boost/integer.hpp>
        from <boost/multiprecision/cpp_int/cpp_int_config.hpp>
    <boost/integer/common_factor_rt.hpp>
        from <boost/multiprecision/cpp_int/misc.hpp>
    <boost/integer/static_min_max.hpp>
        from <boost/multiprecision/cpp_int.hpp>
    <boost/integer_traits.hpp>
        from <boost/multiprecision/cpp_int/cpp_int_config.hpp>

lexical_cast:
    <boost/lexical_cast.hpp>
        from <boost/multiprecision/detail/number_base.hpp>

math:
    <boost/math/constants/constants.hpp>
[...]

PR #373, opened 2021-10-02
Remove dependencies from cpp_int_config

mborland:
Removes uses of boost.assert, and boost.integer. Also implements `BOOST_MP_STANDALONE` to continue work on issue #364.

### DIFF
--- a/include/boost/multiprecision/cpp_int/cpp_int_config.hpp
+++ b/include/boost/multiprecision/cpp_int/cpp_int_config.hpp
@@ -1,17 +1,17 @@
 ///////////////////////////////////////////////////////////////
-//  Copyright 2012 John Maddock. Distributed under the Boost
-//  Software License, Version 1.0. (See accompanying file
-//  LICENSE_1_0.txt or copy at https://www.boost.org/LICENSE_1_0.txt
+//  Copyright 2012 - 2021 John Maddock.
+//  Copyright 2021 Matt Borland.
+//  Distributed under the Boost Software License, Version 1.0.
+//  See accompanying file LICENSE_1_0.txt or copy at https://www.boost.org/LICENSE_1_0.txt
 
 #ifndef BOOST_MP_CPP_INT_CORE_HPP
 #define BOOST_MP_CPP_INT_CORE_HPP
 
+#include <cstdint>
 #include <type_traits>
 #include <limits>
 #include <boost/multiprecision/detail/number_base.hpp>
 #include <boost/multiprecision/detail/assert.hpp>
-#include <boost/integer.hpp>
-#include <boost/integer_traits.hpp>
 
 namespace boost {
 namespace multiprecision {
@@ -21,11 +21,34 @@ namespace detail {
 //
 // These traits calculate the largest type in the list
 // [unsigned] long long, long, int, which has the specified number
-// of bits.  Note that intN_t and boost::int_t<N> find the first
+// of bits.  Note that int_t and uint_t find the first
 // member of the above list, not the last.  We want the last in the
 // list to ensure that mixed arithmetic operations are as efficient
 // as possible.
 //
+
+template <unsigned Bits>
+struct int_t
+{
+   using exact = typename std::conditional<Bits == sizeof(signed char) * CHAR_BIT, signed char,
+                 typename std::conditional<Bits == sizeof(short) * CHAR_BIT, short,
+                 typename std::conditional<Bits == sizeof(int) * CHAR_BIT, int,
+                 typename std::conditional<Bits == sizeof(long) * CHAR_BIT, long, long long
+                 >::type>::type>::type>::type;
+
+};
+
+template <unsigned Bits>
+struct uint_t
+{
+   using exact = typename std::conditional<Bits == sizeof(unsigned char) * CHAR_BIT, unsigned char,
+                 typename std::conditional<Bits == sizeof(unsigned short) * CHAR_BIT, unsigned short,
+                 typename std::conditional<Bits == sizeof(unsigned int) * CHAR_BIT, unsigned int,
+                 typename std::conditional<Bits == sizeof(unsigned long) * CHAR_BIT, unsigned long, unsigned long long
+                 >::type>::type>::type>::type;
+
+};
+
 template <unsigned N>
 struct largest_signed_type
 {
@@ -38,7 +61,7 @@ struct largest_signed_type
            typename std::conditional<
                1 + std::numeric_limits<int>::digits == N,
                int,
-               typename boost::int_t<N>::exact>::type>::type>::type;
+               typename int_t<N>::exact>::type>::type>::type;
 };
 
 template <unsigned N>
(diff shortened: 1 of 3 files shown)


include/boost/multiprecision/cpp_int/cpp_int_config.hpp
```diff
@@ -17,57 +20,80 @@ namespace detail {
...
+               typename uint_t<N>::exact>::type>::type>::type;
 };
 
 } // namespace detail
 
 #if defined(BOOST_HAS_INT128)
 
 using limb_type = detail::largest_unsigned_type<64>::type;
-using signed_limb_type = detail::largest_signed_type<64>::type  ;
-using double_limb_type = boost::uint128_type                    ;
```
jzmaddock, line 62: Using these typedefs suppresses endless warnings from the compiler about using a non-std extension - perhaps we can lift that logic from Boost.Config?

mborland: That should be easy enough. I think I need to define `BOOST_HAS_INT128` in a standalone config like math has.

mborland:
@jzmaddock The only test failure was a timeout. Should be good pending further review.

include/boost/multiprecision/cpp_int/cpp_int_config.hpp
```diff
@@ -17,57 +21,80 @@ namespace detail {
...
 // as possible.
 //
+
+template <unsigned Bits>
+struct int_t
+{
+   using exact = typename std::conditional<Bits == sizeof(signed char) * CHAR_BIT, signed char,
+                 typename std::conditional<Bits == sizeof(short) * CHAR_BIT, short,
+                 typename std::conditional<Bits == sizeof(int) * CHAR_BIT, int,
+                 typename std::conditional<Bits == sizeof(long) * CHAR_BIT, long, long long
```
jzmaddock, line 37: Out of an over-abundance of caution, there's just a couple more things we should do here I think.

In this case, before setting `type` to `[unsigned] long long`, we should check that the bit count matches up, in the case that we have no matching type, maybe the best we can do is set exact to to type `void`, any better ideas?

include/boost/multiprecision/detail/assert.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <boost/multiprecision/detail/standalone_config.hpp>
+
+#ifndef BOOST_MP_STANDALONE
+
+#include <boost/assert.hpp>
+#include <boost/static_assert.hpp>
+#define BOOST_MP_ASSERT(expr) BOOST_ASSERT(expr)
+#define BOOST_MP_ASSERT_MSG(expr, msg) BOOST_ASSERT_MSG(expr, msg)
+#define BOOST_MP_STATIC_ASSERT(expr) BOOST_STATIC_ASSERT(expr)
```
jzmaddock, line 21: I don't think these are in use any more - can we check and maybe remove all references to BOOST_STATIC_ASSERT?

include/boost/multiprecision/detail/standalone_config.hpp
```diff
@@ -0,0 +1,71 @@
...
+#ifdef (__SIZEOF_INT128__)
+#  define BOOST_HAS_INT128
+#endif
+
+#ifdef (__SIZEOF_FLOAT128__)
+#  define BOOST_HAS_FLOAT128
+#endif
+
+#if defined(BOOST_HAS_INT128) && defined(__cplusplus)
+namespace boost{
```
jzmaddock, line 47: I think we should avoid putting these directly in namespace boost: imagine 2 different Boost libraries doing the same thing.  Suggest we put these directly into boost::multiprecision here, and remove the definitions in cpp_int_config.hpp (or vice versa?)

include/boost/multiprecision/detail/standalone_config.hpp
```diff
@@ -0,0 +1,71 @@
...
+#else // Standalone mode
+
+// Prevent Macro sub
+#ifndef BOOST_PREVENT_MACRO_SUBSTITUTION
+#  define BOOST_PREVENT_MACRO_SUBSTITUTION
+#endif
+
+// int/float 128 types
+
+#ifdef (__SIZEOF_INT128__)
```
jzmaddock, line 38: This will more or less work for vanilla GCC, but if you look at Boost.Config the ways in which this gets defined (or undefined) are pretty complex.  Not sure if we need to worry too much - maybe the answer is "use Boost" if the configuration is wrong... only SNAFU is that everything may compile but be inefficient if __int128 wasn't enabled.

jzmaddock:
@NAThompson , @ckormanyos , 

Just thinking out loud here, but where are we going with standalone usage?

For example: a user just needing mpz_int probably has very few dependencies already, someone using cpp_Int might reasonably be required to have Boost.Config available, and for tgamma with cpp_bin_float we would need Boost.Math as well.  So one definition of "standalone" might be zero dependencies unless you use feature "X" in which case library "Y" is required.

So we could:

* For "difficult" configuration decisions, require Boost.Cong, this would be a #error along with a helpful message if not available.
* For functions implemented by Boost.Math we would make instantiating those functions generate static_assert's unless the Math library is available.

What do you all think?  Am I over-thinking this?

NAThompson:
> Just thinking out loud here, but where are we going with standalone usage?

Is it possible to do standalone config + multiprecision? I think nobody is gonna complain about that dependency set.

mborland:
> > Just thinking out loud here, but where are we going with standalone usage?
> 
> Is it possible to do standalone config + multiprecision? I think nobody is gonna complain about that dependency set.

config + multiprecision wouldn't be overly difficult since config currently has no dependencies. I like @jzmaddock suggestion of offering vanilla (e.g. x86_64 + GCC + Linux) config, but then something more difficult (e.g. z/Architecture) requiring config. Would this fit your use case @NAThompson?

NAThompson:
> Would this fit your use case @NAThompson?

Yes. Actually, a config + multiprecision would be fine as well.

mborland:
> Yes. Actually, a config + multiprecision would be fine as well.

Since config is already dependency free is it possible to add it a submodule/link like the [Microsoft STL has for Boost.Math](https://github.com/microsoft/STL)? That would make packaging the dependency easy, and should be totally transparent to the end user.

NAThompson:
> Since config is already dependency free is it possible to add it a submodule/link like the Microsoft STL has for Boost.Math? 

That's a good idea, IMO. I'm not quite sure how the mechanics work, esp. if the entire boost is also on the syspath.

jzmaddock:
>Since config is already dependency free is it possible to add it a submodule/link like the Microsoft STL has for Boost.Math? That would make packaging the dependency easy, and should be totally transparent to the end user.

I don't think we can formally add it as a git submodule, but we could package it up as part of the download if we do independent releases?

mborland:
> I don't think we can formally add it as a git submodule, but we could package it up as part of the download if we do independent releases?

That makes sense too. Offer something like a tarball, a `.deb`, and a `.rpm`?

NAThompson:
>  Offer something like a tarball, a .deb, and a .rpm?

Github by default zips up your code into a zip file. I think we'd need to prepare a zip file with both config and multiprecision, then make sure config is properly hidden with a `CMakeLists.txt` which is configured correctly and doesn't break the main project.

I imagine this will be 10 lines of code, but extremely difficult to get right.

mborland:
This latest iteration is clean. The only CI failure is a timeout.

PR #379, opened 2021-10-09
Remove boost.predef dependency

mborland:


include/boost/multiprecision/detail/endian.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#  include <boost/predef/other/endian.h>
+#  define BOOST_MP_ENDIAN_BIG_BYTE BOOST_ENDIAN_BIG_BYTE
+#  define BOOST_MP_ENDIAN_LITTLE_BYTE BOOST_ENDIAN_LITTLE_BYTE
+
+#elif (__cplusplus >= 202002L || _MSVC_LANG >= 202002L)
+
+#  if __has_include(<bit>)
+#    include <bit>
+#    define BOOST_MP_ENDIAN_BIG_BYTE (std::endian::native == std::endian::big)
```
jzmaddock, line 21: Sadly, I don't believe this will work since the types and values used here are not usable in a preprocessor conditional?

include/boost/multiprecision/cpp_int/import_export.hpp
```diff
@@ -6,6 +6,10 @@
 #ifndef BOOST_MP_CPP_INT_IMPORT_EXPORT_HPP
 #define BOOST_MP_CPP_INT_IMPORT_EXPORT_HPP
 
+#include <climits>
```
jzmaddock, line 9: Just curious, why the extra #include's here and elsewhere, were we getting these accidentally via predef/endian?

include/boost/multiprecision/detail/endian.hpp
mborland: You are correct. `std::endian` is just an `enum` aliasing the byte order macros (with libstdc++ anyway).

mborland:
Here and in Math CircleCI says running is paused due to concurrency limits. Did we run out of credits again?

PR #380, opened 2021-10-10
Remove boost.random dependency

mborland:
In standalone mode the functionality of `miller_rabin_test` is reduced to only those types that are compatible with `std::uniform_int_distribution`.

### DIFF
--- a/example/safe_prime.cpp
+++ b/example/safe_prime.cpp
@@ -5,6 +5,7 @@
 
 //[safe_prime
 
+#include <boost/random.hpp>
 #include <boost/multiprecision/cpp_int.hpp>
 #include <boost/multiprecision/miller_rabin.hpp>
 #include <iostream>
--- a/include/boost/multiprecision/detail/uniform_int_distribution.hpp
+++ b/include/boost/multiprecision/detail/uniform_int_distribution.hpp
@@ -42,7 +42,7 @@ struct make_unsigned_mp
     using type = typename make_unsigned_impl<T, boost::multiprecision::detail::is_integral<T>::value>::type;
 };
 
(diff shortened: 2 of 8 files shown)


ckormanyos:
> In standalone mode the functionality of `miller_rabin_test` is reduced to only those types that are compatible with `std::uniform_int_distribution`

> Porting from Boost.Random would be possible if needed, 

I think it will be quite a relief to remove the dependency on Boost.Random.

I'm not sure, but thinking out loud here. There are tests and at least one example that require wide integer types for testing Miller-Rabin. But basically these tests simple need a random uniform bit-stream genertor for N bits. With just a few lines of C++11 code, it should be possible to make or fashion a little utility from an integer distribution on the range (0,1) for N bits that could then be used in these tests and examples --- standalone or not. I can help if this is thought to be feasible.

ckormanyos:
> have ported `uniform_int_distribution` and modernized the implementation to use C++11 internally removing any external dependencies. This breaks the dependency on Boost.Random with or without standalone mode.

I think it is a good detail/utility that removes a rather heavy dependency.

mborland:
CI on this one is clean so pending review it should be good to go.

mborland:
This is clean with the requested changes. Only failure is internal to the CI.

ckormanyos:
Hi @mborland and @jzmaddock this is cool.

In another project I just experienced a problem because my project had (falsely) relied on the parasitic inclusion of some `Boost.Random` headers via the inclusion of `<boost/multiprecision/miller_rabin.hpp>`.

Although this is/was a problem in my particular project, it wouldn't surprise me if others experience such things in 1.79 down the road. Is it important for others? Can we announce explicitly in changes, as I think a few others might run into this nuance.

PR #394, opened 2021-11-22
Remove boost.integer dependency

mborland:


include/boost/multiprecision/detail/common_factor.hpp
```diff
@@ -0,0 +1,65 @@
...
+    return boost::integer::lcm(a, b);
+}
+}}} // Namespaces
+
+#else 
+
+namespace boost { namespace multiprecision { namespace detail { 
+
+template <typename T>
+inline BOOST_CXX14_CONSTEXPR void constexpr_swap(T& a, T& b) noexcept
```
jzmaddock, line 36: I wonder/worry a bit about this - gcd is typically time-critical and binary-gcd is usually faster than Euclid.  We also already have: https://github.com/boostorg/multiprecision/blob/d6b0f2ecd50992e868652806e437991c771dc6b2/include/boost/multiprecision/cpp_int/misc.hpp#L397-L437 and perhaps we should do something similar here and/or share code?

mborland:
This is clean with the requested changes.

PR #404, opened 2021-12-15
Change all bit counts to use size_t/ptrdiff_t

jzmaddock:


ckormanyos:
Wow --- strong and courageous. Thank you John!

PR #409, opened 2022-01-02
Standalone fixes

mborland:
The diff stat on this one will be ugly because it reverts @ckormanyos reverting the merge of PR #403. Fixes the review comments from @jzmaddock that led to the revert:

1. Fixes for Eigen types needing the use of `boost::math::tools` instead of `std::numeric_limits` functions.
2. Replace the use of `std::free` with `gmp::free_function` for better interoperability.


jzmaddock:
Just a heads up that this PR https://github.com/boostorg/multiprecision/pull/408 contains fixes that will likely clash with this: in particular all the lexical-based conversions in tommath.hpp are removed in favour of "proper" conversion operators.

mborland:
Merged in #408 and went with all of your tommath changes in merge conflicts. Good news is there were not many.

jzmaddock:
>Merged in #408 and went with all of your tommath changes in merge conflicts. Good news is there were not many. Once your PR is merged to develop I can rebase to clean this up.

#408 is merged now, can you rebase so we can see a cleaner change set?

Many thanks for persevering with this!

include/boost/multiprecision/gmp.hpp
```diff
@@ -3734,35 +3801,39 @@ namespace Eigen
          IsSigned = std::numeric_limits<self_type>::is_specialized ? std::numeric_limits<self_type>::is_signed : true,
          RequireInitialization = 1,
       };
+
+      #if !defined(BOOST_MP_STANDALONE) || defined(BOOST_MATH_STANDALONE)
       static Real epsilon()
       {
-         return boost::math::tools::epsilon<Real>();
+         return std::numeric_limits<Real>::epsilon();
```
jzmaddock, line 3808: This still doesn't look quite right to me: if that PP condition is met, then Boost.Math is available, and we can (and should) be using the Boost.Math functions?

jzmaddock: On second thoughts.... this is a traits-class specialization for gmp_float<0> only, and all the boost math functions except epsilon are defined right here in this file, and epsilon is just `ldexp(Real(1), 1 - digits())`, so the #if logic can go, and we can call the boost::math functions which are already there and do the right thing?

jzmaddock:
Other than the comments above, this all looks good to me - many many thanks for this Matt - much appreciated!

mborland:
I think this is good to go now. The drone run is hung; some of the test have been "running" for over 9 hours.

ckormanyos:
> I think this is good to go now.

This is really great. Thank you Matt! Thanks for also guiding this endeavor, John.

@jzmaddock I'm very, very happy with standalone and will leave this merge to you, just in case you've got something else going on.

When the dust settles, I might swing in for some trivial embedded bare-metal tweaks on high-level warnings and string-streaming, but those can wait indefinitely.

jzmaddock:
Looks good, merging...

PR #412, opened 2022-01-15
Check for over large reduction before we assert.

jzmaddock:
Found why testing sinh_sinh integration in Boost.Math.


ckormanyos:
Wow! What a hard-to-find program artifact. Masterful catch, John!

PR #417, opened 2022-01-22
Fix some mpq rational arithmetic errors:

jzmaddock:
Multiplication by zero should not proceed via gcd.
Multiplication by scalar should be restricted to integer types (not floats).
Add some more tests to catch these cases.

### DIFF
--- a/include/boost/multiprecision/gmp.hpp
+++ b/include/boost/multiprecision/gmp.hpp
@@ -2754,19 +2754,19 @@ inline void eval_add(gmp_rational& result, gmp_rational const& a, long b)
    // no need to normalize, there can be no common divisor as long as a is already normalized.
 }
 template <class T>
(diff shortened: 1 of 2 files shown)


afabri:
Thank you very much.

afabri:
What is the process and schedule for fixes in the mutiprecision package to make it into the master branch of boost and into a boost release ?

jzmaddock:
> What is the process and schedule for fixes in the mutiprecision package to make it into the master branch of boost and into a boost release ?

At some point before the next release we'll merge everything to master, I can't find the release schedule off hand, but Easter sometime I would expect the next Boost release to drop.

afabri:
@jzmaddock do you get a kind of notification when this fix makes it to the boost master branch (or is it even you who merges)? If so could you do me a favor and post it in this conversation. Thanks, Andreas

PR #418, opened 2022-01-23
Math standalone

mborland:
Offer better context detection of Boost.Math standalone in conjunction with Boost.MP standalone.

### DIFF
--- a/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
+++ b/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
@@ -14,9 +14,10 @@
 #include <iomanip>
 #include <tuple>
 #include <functional>
+#include <boost/multiprecision/detail/standalone_config.hpp>
 #include <boost/multiprecision/number.hpp>
-#include <boost/math/special_functions/fpclassify.hpp>
 #include <boost/multiprecision/detail/no_exceptions_support.hpp>
+#include <boost/multiprecision/detail/fpclassify.hpp>
 
 namespace boost {
 namespace multiprecision {
@@ -208,7 +209,7 @@ inline void eval_sqrt(number_backend_float_architype& result, const number_backe
 
 inline int eval_fpclassify(const number_backend_float_architype& arg)
 {
-   return (boost::math::fpclassify)(arg.m_value);
+   return (boost::multiprecision::detail::fpclassify)(arg.m_value);
 }
 
 inline std::size_t hash_value(const number_backend_float_architype& v)
--- a/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
+++ b/include/boost/multiprecision/concepts/mp_number_archetypes.hpp
@@ -209,7 +209,7 @@ inline void eval_sqrt(number_backend_float_architype& result, const number_backe
 
 inline int eval_fpclassify(const number_backend_float_architype& arg)
 {
-   return (boost::multiprecision::detail::fpclassify)(arg.m_value);
+   return BOOST_MP_FPCLASSIFY(arg.m_value);
 }
 
 inline std::size_t hash_value(const number_backend_float_architype& v)
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -2430,7 +2430,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
    if (a == 1)
       return *this = one();
 
-   if ((boost::multiprecision::detail::isinf)(a))
+   if (BOOST_MP_ISINF(a))
    {
       *this = inf();
       if (a < 0)
@@ -2438,7 +2438,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
       return *this;
    }
 
-   if ((boost::multiprecision::detail::isnan)(a))
+   if (BOOST_MP_ISNAN(a))
       return *this = nan();
 
    int         e;
@@ -2447,7 +2447,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
 
    f = frexp(a, &e);
    // See https://svn.boost.org/trac/boost/ticket/10924 for an example of why this may go wrong:
-   BOOST_MP_ASSERT(!(boost::multiprecision::detail::isnan)(f) && !(boost::multiprecision::detail::isinf)(f));
+   BOOST_MP_ASSERT(!BOOST_MP_ISNAN(f) && !BOOST_MP_ISINF(f));
 
    constexpr const int shift = std::numeric_limits<int>::digits - 1;
 
@@ -2455,7 +2455,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
    {
       // extract int sized bits from f:
       f = ldexp(f, shift);
-      BOOST_MP_ASSERT(!(boost::multiprecision::detail::isnan)(f) && !(boost::multiprecision::detail::isinf)(f));
+      BOOST_MP_ASSERT(!BOOST_MP_ISNAN(f) && !BOOST_MP_ISINF(f));
       term = floor(f);
       e -= shift;
       *this *= pow2(shift);
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -3587,6 +3587,7 @@ constexpr float_round_style numeric_limits<boost::multiprecision::number<boost::
 
 } // namespace std
 
+#ifdef BOOST_MP_MATH_AVAILABLE
 namespace boost {
 namespace math {
 
@@ -3612,6 +3613,7 @@ struct precision<boost::multiprecision::number<boost::multiprecision::cpp_dec_fl
 }
 
 }} // namespace boost::math::policies
+#endif
 
 #ifdef BOOST_MSVC
 #pragma warning(pop)
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -2430,7 +2430,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
    if (a == 1)
       return *this = one();
 
-   if (boost::multiprecision::detail::isinf(a))
+   if ((boost::multiprecision::detail::isinf)(a))
    {
       *this = inf();
       if (a < 0)
@@ -2438,7 +2438,7 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
       return *this;
    }
 
-   if (boost::multiprecision::detail::isnan(a))
+   if ((boost::multiprecision::detail::isnan)(a))
       return *this = nan();
 
    int         e;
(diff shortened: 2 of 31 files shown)


mborland:
@jzmaddock Almost all of my failures are in the concepts test where `fpclassify`/`isnan`/`isinf` are replaced with a macro that is designed to generate a compilation failure ([example](https://github.com/boostorg/multiprecision/runs/4925863564?check_suite_focus=true#step:17:18982)). Is there a workaround for this or something I am missing? I am just calling the `boost::math` implementation for almost all of these (see new `<boost/multiprecision/detail/fpclassify.hpp>`)

jzmaddock:
Although they're declining in number, there are some platforms that define isnan/isinf etc as macros (rather like the min/max issue on Windows), so all those functions need to be macro expansion proof.

For declarations we can do that with:

```
bool isnan BOOST_PREVENT_MACRO_SUBSTITUTION(T args); // or
bool isnan /*Suppress macro expansion*/(T args);
```

For calls it's either:

```
(boost::math::isnan)(x) // or
boost::math::isnan  BOOST_PREVENT_MACRO_SUBSTITUTION(x); // or
boost::math::isnan /*Suppress macro expansion*/(x);
```

HTH, John.

mborland:
@jzmaddock Can you please kill the current CI run for this? Something seems very wrong in the arithmetic tests and will take 6 hours per test to time out. Thanks.

jzmaddock:
Hi Matt, thanks for persevering with this, I think we can simply things here quite a bit, but I need to check through the changes line by line to be sure, so be prepared for a small blizzard of line-comments!

include/boost/multiprecision/concepts/mp_number_archetypes.hpp
```diff
@@ -208,7 +209,7 @@ inline void eval_sqrt(number_backend_float_architype& result, const number_backe
 
 inline int eval_fpclassify(const number_backend_float_architype& arg)
 {
-   return (boost::math::fpclassify)(arg.m_value);
+   return (boost::multiprecision::detail::fpclassify)(arg.m_value);
```
jzmaddock, line 212: This can just be std::fpclassify here.

include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -2416,23 +2419,26 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
...
+   using std::frexp; 
+   using std::ldexp;
 
    if (a == 0)
       return *this = zero();
 
    if (a == 1)
       return *this = one();
 
-   if ((boost::math::isinf)(a))
```
jzmaddock, line 2427: The function is enable_if'ed on is_floating_point so we can use std::isinf.

include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -2416,23 +2419,26 @@ typename std::enable_if<std::is_floating_point<Float>::value, cpp_dec_float<Digi
...
+   if ((boost::multiprecision::detail::isinf)(a))
```
jzmaddock, line 2433: And std::isnan, and so on down this function.

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -1628,7 +1630,7 @@ struct cpp_int_backend
          *this = static_cast<limb_type>(1u);
       }
 
-      if ((boost::math::isinf)(a) || (boost::math::isnan)(a))
```
jzmaddock, line 1631: Again, only used for builtin floats, so we can use std::isinf etc here.

include/boost/multiprecision/detail/fpclassify.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#else
+
+namespace boost { namespace multiprecision { namespace detail {
+
+// Assume IEEE 754 compliance and compiler optimizations do not affect this
+template <typename T>
+inline constexpr bool isnan BOOST_PREVENT_MACRO_SUBSTITUTION (const T val) noexcept
+{
+    return val != val;
```
jzmaddock, line 47: I'm hoping we can ditch this whole header, but the val != val trick is known not to work for some optimisation levels.  Actually std::isnan breaks when some optimisations are enabled for gcc as well, but we have no control over that :(

include/boost/multiprecision/detail/functions/trunc.hpp
```diff
@@ -0,0 +1,95 @@
...
+    else
+    {
+        return result + 1;
+    }
+}
+
+template <typename T>
+inline T trunc BOOST_PREVENT_MACRO_SUBSTITUTION (const T arg)
+{
+    return (arg > 0) ? boost::multiprecision::detail::impl::floor(arg) : boost::multiprecision::detail::impl::ceil(arg);
```
jzmaddock, line 65: I would tend to go with:

```
using std::floor;  using std::ceil;
return (arg > 0) ? floor(arg) : ceil(arg);
```

include/boost/multiprecision/detail/number_compare.hpp
```diff
@@ -152,7 +153,7 @@ inline
    else
 #endif
    {
-      return (boost::math::isnan)(a);
+      return (boost::multiprecision::detail::isnan)(a);
```
jzmaddock, line 156: I think we could use:

```
using std::isnan;
return isnan /*prevent macro expansion*/(a);
```

include/boost/multiprecision/gmp.hpp
```diff
@@ -300,10 +300,8 @@ struct gmp_float_imp
          return *this;
       }
 
-      #ifndef BOOST_MP_STANDALONE
-      BOOST_MP_ASSERT(!(boost::math::isinf)(a));
-      BOOST_MP_ASSERT(!(boost::math::isnan)(a));
-      #endif
+      BOOST_MP_ASSERT(!(boost::multiprecision::detail::isinf)(a));
```
jzmaddock, line 303: This code has to work with float, double, long double and __float128, I think we can use `isinf` etc as unqualified calls (with macro expansion suppression), and update the using declarations at the start, plus BOOST_MP_FLOAT128_USING to bring the functions into scope.

include/boost/multiprecision/gmp.hpp
```diff
@@ -1521,10 +1519,8 @@ struct gmp_int
...
```
jzmaddock, line 1522: as above.

include/boost/multiprecision/gmp.hpp
```diff
@@ -2509,10 +2505,8 @@ struct gmp_rational
...
```
jzmaddock, line 2508: as above.

include/boost/multiprecision/mpc.hpp
```diff
@@ -1370,11 +1372,11 @@ inline void assign_components(mpc_complex_backend<D1>& result, long long a, long
 template <unsigned D1, unsigned D2, mpfr_allocation_type AllocationType>
 inline void assign_components(mpc_complex_backend<D1>& result, double a, double b)
 {
-   if ((boost::math::isnan)(a))
+   if ((boost::multiprecision::detail::isnan)(a))
```
jzmaddock, line 1375: We can use std::isnan etc throughout here, as it's all double or long double.

include/boost/multiprecision/mpfi.hpp
```diff
@@ -284,8 +284,8 @@ struct mpfi_float_imp
          return *this;
       }
 
-      BOOST_MP_ASSERT(!(boost::math::isinf)(a));
-      BOOST_MP_ASSERT(!(boost::math::isnan)(a));
+      BOOST_MP_ASSERT(!(boost::multiprecision::detail::isinf)(a));
```
jzmaddock, line 287: Argument is __float128, as before update BOOST_MP_FLOAT128_USING and make an unqualified call.

include/boost/multiprecision/mpfr.hpp
```diff
@@ -317,8 +323,8 @@ struct mpfr_float_imp<digits10, allocate_dynamic>
          return *this;
       }
 
-      BOOST_MP_ASSERT(!(boost::math::isinf)(a));
-      BOOST_MP_ASSERT(!(boost::math::isnan)(a));
+      BOOST_MP_ASSERT(!(boost::multiprecision::detail::isinf)(a));
```
jzmaddock, line 326: Another where an update to BOOST_MP_FLOAT128_USING and unqualified call fixes things.

include/boost/multiprecision/tommath.hpp
```diff
@@ -245,10 +243,8 @@ struct tommath_int
          return *this;
       }
 
-      #if !defined(BOOST_MP_STANDALONE) || defined(BOOST_MATH_STANDALONE)
-      BOOST_MP_ASSERT(!(boost::math::isinf)(a));
-      BOOST_MP_ASSERT(!(boost::math::isnan)(a));
-      #endif
+      BOOST_MP_ASSERT(!(boost::multiprecision::detail::isinf)(a));
```
jzmaddock, line 246: Another where an unqualified call and update to BOOST_MP_FLOAT128_USING would fix things.

jzmaddock:
OK, apologies for the comment blizzard, but I think we can simplify things a lot with this.

Note that in general, I'd rather not replace calls to boost::math::isnan to std::isnan as the former keeps working even when certain "fast-and-loose" optimisations are in place.  One option to reduce the #if logic, would be a macro that expands to either `using std::isnan` or `using boost::math::isnan` depending on which is available.

mborland:
I think this latest commit simplifies what you were looking for. It defines macros such as `BOOST_MP_ISINF` that either expands to `boost::math::isinf` if it is available, or SFINAEd definitions of `isinf` depending on if the type is supported by existing definitions (e.g. STL and libquadmath)

jzmaddock:
Thanks Matt, can we just get rid of the floor/ceil code in place of std::floor/ceil, and then hopefully this should be good to go?

mborland:
> Thanks Matt, can we just get rid of the floor/ceil code in place of std::floor/ceil, and then hopefully this should be good to go?

Locally I don't see any issues with std::floor/ceil so the superfluous code is gone now. Sorry I missed that one with the fixes earlier.

mborland:
@jzmaddock This should be good now. The 2 failures are timeouts on the CI's side.

mborland:
Upon a deeper look this is in fact not good to go. I spun up a VM that has nothing installed to see where we are at. There are more lingering instances that need to be removed/replaced.

mborland:
@jzmaddock This is now good to go. I pulled this branch into a VM with none of boost installed and it all compiles now.

mariospr:
FWIW, and just in case it helps: I was initially not able able to build this multiprecision library in standalone mode off the `develop` branch because of several references to `<boost/math/...>` not guarded by the `BOOST_MP_STANDALONE` define, when I tried to do it yesterday for the first time before finding this PR.

But then I found this PR and I can confirm I could build in standalone mode now with the only addition of Boost.Config, which I had to download as well as it's still referenced from standalone_config.hpp (despite of the [comment saying that this package _"already includes a copy of Boost.Config so no other donwloads are required_"](github.com/boostorg/multiprecision/pull/418), which I must say confused me a bit).

Anyway, this is not a review (I'm not qualified at all to do so), just a quick comment based on my recent experience in case it helps in some way. Thanks a lot for this PR in any case!.

mborland:
> 

@mariospr I am glad to hear that this PR works for you. The intent is to have a packaged standalone release on the home page much like we do with [Boost.Math](https://github.com/boostorg/math/releases) that will include Boost.Config. The updates to the docs may have been pre-mature so I apologize for the confusion.

mariospr:
> @mariospr I am glad to hear that this PR works for you. The intent is to have a packaged standalone release on the home page much like we do with [Boost.Math](https://github.com/boostorg/math/releases) that will include Boost.Config. 

Ah, I see, that makes a lot of sense, thanks for clarifying.

> The updates to the docs may have been pre-mature so I apologize for the confusion.

No worries. I'm not too familiar with Boost so I was unsure after reading the README whether I was missing something out, but this explains things, thanks again.

mborland:
Ping @jzmaddock. We are coming up on the beta/release window, and I think it's important this makes it.

PR #430, opened 2022-02-04
Collected fixes for issue 427

jzmaddock:


jzmaddock:
Has something changed in the Github visual studio images?  All the msvc tests are failing to find msvc: https://github.com/boostorg/multiprecision/runs/5071594639?check_suite_focus=true

jzmaddock:
Ah, we were using windows-latest for the msvc tests, and the meaning of that has just changed....

PR #432, opened 2022-02-05
[ci skip] Fixes #431 via not-defined __clang__

### LINKED ISSUE #431: -Wmaybe-uninitialized clang in pragma in cpp_int.hpp
In our recent correction to eliminate bogus warning on uninitialized, we inserted in `#pragma`s in the attempted suppression of `-maybe-uninitialized`.

This results in an new unknown warning on `clang++` which reports that it is not aware or the meaning of this warning suppression.

So in lines like [here](https://github.com/boostorg/multiprecision/blob/28b314828ddd06b3e2e3586758b398fc29ceeb89/include/boost/multiprecision/cpp_int.hpp#L39), I think we need to test for:

```cpp
#if defined(__GNUC__) && !defined(__clang__)
// Suppressioin stuff
#endif
```

ckormanyos:


jzmaddock:
Looks good, this PR (https://github.com/boostorg/multiprecision/pull/430) fixes some CI issues that have just cropped up (among other things), once that's green and in we can merge this and other waiting PR's.

ckormanyos:
>  once that's green and in we can merge this and other waiting PR's

Understood. Thanks john.

Issue #434, opened 2022-02-06
More GCC warnings


jzmaddock:
When compiling test_arithmetic_cpp_bin_float_2m.cpp with:

```
g++ -O3 -DNDEBUG -c test_arithmetic_cpp_bin_float_2m.cpp
```

 I see:

```
In file included from ../../../boost/multiprecision/cpp_int.hpp:2338,
                 from ../../../boost/multiprecision/cpp_bin_float.hpp:11,
                 from test_arithmetic_cpp_bin_float_2m.cpp:8:
In function 'constexpr void boost::multiprecision::backends::left_shift_generic(Int&, boost::multiprecision::double_limb_type) [with Int = boost::multiprecision::backends::cpp_int_backend<0, 33
24, boost::multiprecision::signed_magnitude, boost::multiprecision::unchecked, std::allocator<char> >]',
    inlined from 'constexpr typename std::enable_if<(! boost::multiprecision::backends::is_trivial_cpp_int<boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator> >::value)>::type boost::multiprecision::backends::eval_left_shift(boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>&, boost::multiprecision::double_limb_type) [with long long unsigned int MinBits1 = 0; long long unsigned int MaxBits1 = 3324; boost::multiprecision::cpp_integer_type SignType1 = boost::multiprecision::signed_magnitude; boost::multiprecision::cpp_int_check_type Checked1 = boost::multiprecision::unchecked; Allocator1 = std::allocator<char>]' at ../../../boost/multiprecision/cpp_int/bitwise.hpp:493:25,
[...]

PR #435, opened 2022-02-07
Use an explicit memset in bitwise.hpp shift operators.

jzmaddock:
Rather than relying on compiler generated transformation to memset, as the latter results in spurious GCC warnings.
Fixes https://github.com/boostorg/multiprecision/issues/434.


jzmaddock:
Failing test is a network timeout, merging...

PR #440, opened 2022-03-06
Consistently use BOOST_MSCV to guard warning(push/pop) pragmas

mariospr:
There was some inconsistency as in some places _MSC_VER was used
instead, plus there was a bug on cpp_int.hpp where BOOST_MSVC was
used for warning(push) while _MSC_VER was used for warning(pop).


jzmaddock:
Thanks @mariospr , in an over-abundance of caution, I'll let the CI checks finish and then merge.

mariospr:
Cool, thanks!

Issue #445, opened 2022-04-06
Harden code with higher GCC warnings


ckormanyos:
The purpose of this issue is to raise awarenes for possible code hardening via use of stricter/higher GCC warnings.

Consider starting with the set:

```sh
g++ -Wall -Wconversion -Wextra -Wpedantic -Wshadow -Wundef
```

Further details such as potential CI job(s), how to best detect warning instances, etc. are partly still TODO.

ckormanyos:
First evaluation reveals there is expected to be problems using `-Wpedantic` in association with GCC's non-ANSI data type `__int128`. Either `-Wpedantic` is not appropriate, or some kind of commenting out warnings is needed (which is not pretty because the library actually uses 128 bit types).

ckormanyos:
A good starting point might be the following (on GCC/clang)

```
-Wall -Wextra -Wconversion -Wsign-conversion -Wshadow -Wundef
```

jzmaddock:
If we always use __int128 via a typedef marked with __extension__ (cf __float128 usage), does that fix things?

ckormanyos:
> If we always use `__int128` via a typedef marked with **extension** (cf __float128 usage), does that fix things?

Good questioon John (@jzmaddock). That will be part of this investigation/hardening step. We will attempt to find out if there is a way to secure the reliable use of both `__int128` as well as `__float128` without requiring excessive amounts of code-modification.

I'll look into this as part of this endeavor.

ckormanyos:
So I started the ball rolling on this in a new [enhanced_gcc_clang_warnings](https://github.com/boostorg/multiprecision/tree/enhanced_gcc_clang_warnings) branch.

I created a file called [example/standalone_bernoulli_tgamma.cpp](https://github.com/boostorg/multiprecision/blob/enhanced_gcc_clang_warnings/example/standalone_bernoulli_tgamma.cpp) that can use different MP types.
- Calculate a vector of Bernoulli numbers (not using Math)
- Use Stirling's Approximation to calculate some Gamma function values (not using Math).
- Compare control values.
- The file can either use/not-use `<boost/math/constants/constants.hpp>` with the flag `EXAMPLE008_BERNOULLI_USE_LOCAL_PI` located [here](https://github.com/boostorg/multiprecision/blob/4052c420ac319ae92edec69f656f697644550174/example/standalone_bernoulli_tgamma.cpp#L17).

So when I compile locally with:

```
[...]

PR #447, opened 2022-04-21
Enhanced gcc clang warnings

ckormanyos:
The purpose of this PR is to track handling enhanced GCC/clang warnings as described in #445


ckormanyos:
GHA now allows re-run of failed jobs. Re-runs went green on GHA.

ckormanyos:
The next batch of commits uses the simple test file to get warn-free for bin-float on:
- GCC/VC on MinGW, Win*, WSL
- clang++ on WSL still has 30 warns

Lots of little steps to do. Many of them, in fact. But the method(s) seem feasible and standalone code(s) make this hardening now at least possible.

If/when this goes green, do next:

- Handle clang warns.
- Go for warn-free on gmp-float, mpfr-float, dec-float.
- Extend the test program(s) to cover more code instantiation.
- Set up a companion CI system (maybe a repo) to go hunting for `-Werror`.

Cc: @mborland and @jzmaddock and @NAThompson

ckormanyos:
> whatever

I made a little repo [here](https://github.com/ckormanyos/boost_multiprecision_harden) which pulls in `Config` from the submodule, but carries its own (copied, not cloned) version of `Multiprecision` standalone.

CI in this little repo makes one single test now, but will add a few more to instantiate more stuff as progressing...

Inch forward in this repo, manually merge to the _real_ Boost submod. This kind of thing seems feasible and could also (much later) provide a playground for CodeSonar, Synopsis Coverity, LGTM if we ever decide to go that way.

ckormanyos:
While the next batch of edits cycles I could use some advice.

First off:
- Some non-ASCII chars got into the slash-alsah-slashes near the tops of four headers. I was unsure how they got there and wondered if anyone knows of a Git or MSVC edit setting known to cause this kind of thing.

There are three warnings that I could not handle fixing by myself, or that I would like advice on.

If you get a chance, could John (@jzmaddock) or Matt (@mborland) could you please review Some warnings below?

- On [this line](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_bin_float.hpp#L411), clang [says](https://github.com/ckormanyos/boost_multiprecision_harden/runs/6138663779?check_suite_focus=true#step:4:22) warning: implicit conversion turns floating-point number into integer: `float` to `bool`. Since this is done in the presence of `std::is_floating_point<Float>`, I was thinking comparison to be greater than `numeric_limits<Float>::epsilon()` might be OK. I'd like to avoid comparison with literal floating-point zero as well.
- On [this line](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_int/misc.hpp#L146), clang [says](https://github.com/ckormanyos/boost_multiprecision_harden/runs/6138663779?check_suite_focus=true#step:4:34) warning: implicit conversion changes signedness. I am having difficulty figuring out what type(s) `R` is and how (or if) to cast?
- On exactly the [line that follows](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_int/misc.hpp#L147), clang says warning: implicit conversion changes signedness and I'm again confused how/if to cast?

jzmaddock:
> On [this line](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_bin_float.hpp#L411), clang [says](https://github.com/ckormanyos/boost_multiprecision_harden/runs/6138663779?check_suite_focus=true#step:4:22) warning: implicit conversion turns floating-point number into integer: float to bool. Since this is done in the presence of std::is_floating_point<Float>, I was thinking comparison to be greater than numeric_limits<Float>::epsilon() might be OK. I'd like to avoid comparison with literal floating-point zero as well.

Grrr, the code is correct as is - which is to say (implicitly) "f is not zero".  So if we change it would have to be to `f != 0`.

>On [this line](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_int/misc.hpp#L146), clang [says](https://github.com/ckormanyos/boost_multiprecision_harden/runs/6138663779?check_suite_focus=true#step:4:34) warning: implicit conversion changes signedness. I am having difficulty figuring out what type(s) R is and how (or if) to cast?

If you look at the enable_if above, R is any integral type (signed or unsigned.  I wonder though if I have the brackets in the wrong place here, what about:

```
static_cast<R>((backend.limbs()[i] & mask) << shift)
```
?

>On exactly the [line that follows](https://github.com/boostorg/multiprecision/blob/e8ef5648f46fffd95ff5f45f19c20c82983d27f9/include/boost/multiprecision/cpp_int/misc.hpp#L147), clang [says](https://github.com/ckormanyos/boost_multiprecision_harden/runs/6138663779?check_suite_focus=true#step:4:64) warning: implicit conversion changes signedness and I'm again confused how/if to cast?

Try removing the static_cast to R, it looks superfluous to me, we should just be checking the bits in the limb (in fact this one does look like a possible bug!).

ckormanyos:
Thanks John.

> So if we change it would have to be to `f != 0`

I decided to actually change and went with `f != static_cast<Float>(0.0F)`. There was anothe similar instance I found in `cpp_int.hpp` as well.

> brackets in the wrong place here...

> (in fact this one does look like a possible bug!)

Let's call it a near-bug. i don't really think it's finding a nugget. Anyway, yes. So I went with some [refactoring](https://github.com/boostorg/multiprecision/blob/2c845465baccb7532573a3d92f8e8e238e563d21/include/boost/multiprecision/cpp_int/misc.hpp#L145) around those two lines.

The next batch of changes are cycling. These (in total) handle quite a few (around 75) warnings that I can find locally on my test code and in my own CI of my wide-integer project. In this way, `cpp_dec_float.hpp` and `cpp_int.hpp` have been hardened to some extent. If CI runs through green, I'll commit this first rudimentary step to develop.

ckormanyos:
Hi @jzmaddock this PR addresses about 125 warnings (mostly `-Wconversion` and `-Wsign-conversion` in `cdd_bin_float.hpp`, `cpp_int.hpp` and several headers included by them.

CI ran green, as die all my local tests and i found no performance influence.

I'd like to merge these now as the first small step toward code hardning. When continuing, I'd expect a few hundred more similar edits in all of Multiprecision.

Anyway, thoughts on this first step? I'd like to merge? OK by you?

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -231,11 +233,11 @@ struct cpp_int_base<MinBits, MaxBits, signed_magnitude, Checked, Allocator, fals
 
       constexpr data_type() noexcept : first(0) {}
       constexpr data_type(limb_type i) noexcept : first(i) {}
-      constexpr data_type(signed_limb_type i) noexcept : first(i < 0 ? static_cast<limb_type>(boost::multiprecision::detail::unsigned_abs(i)) : i) {}
+      constexpr data_type(signed_limb_type i) noexcept : first(i < 0 ? static_cast<limb_type>(boost::multiprecision::detail::unsigned_abs(i)) : static_cast<limb_type>(i)) {}
```
jzmaddock, line 236: I'm thinking we could simplify this, and just initialize with the result of `unsigned_abs(i)` which would handle both the `i < 0` check and the cast to limb_type for us?

Same with the next one below.

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -801,7 +803,8 @@ struct cpp_int_base<MinBits, MinBits, unsigned_magnitude, Checked, void, false>
        : m_wrapper(i),
          m_limbs(1) {}
    BOOST_MP_FORCEINLINE BOOST_MP_CXX14_CONSTEXPR cpp_int_base(signed_limb_type i) noexcept((Checked == unchecked))
-       : m_wrapper(limb_type(i < 0 ? static_cast<limb_type>(-static_cast<signed_double_limb_type>(i)) : i)), m_limbs(1)
+       : m_wrapper(i < 0 ? static_cast<limb_type>(-static_cast<signed_double_limb_type>(i)) : static_cast<limb_type>(i)),
```
jzmaddock, line 806: This would be safer using `unsigned_abs` here as well.

include/boost/multiprecision/cpp_int/misc.hpp
```diff
@@ -208,10 +211,14 @@ eval_convert_to(R* result, const cpp_int_backend<MinBits1, MaxBits1, SignType1,
          if (bits_to_keep < (std::ptrdiff_t)cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>::limb_bits)
          {
             if(index != backend.size() - 1)
-               mask <<= cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>::limb_bits - bits_to_keep;
+            {
+               const std::ptrdiff_t left_shift_amount = static_cast<std::ptrdiff_t>(static_cast<std::ptrdiff_t>(cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>::limb_bits) - bits_to_keep);
+
+               mask = static_cast<limb_type>(mask << left_shift_amount);
```
jzmaddock, line 217: Can we really not use <<= here?  It's fine, but it's just harder to read...

jzmaddock:
Other than some very minor quibbles, that all looks fine to me.

include/boost/multiprecision/cpp_int/misc.hpp
ckormanyos: > Can we really not use <<= here? It's fine, but it's just harder to read...

Alas the verbosity of my lines is scaling with my increasing age. Sure. I can terse-ify.

include/boost/multiprecision/cpp_int.hpp
ckormanyos: Done

include/boost/multiprecision/cpp_int.hpp
ckormanyos: Done

jzmaddock:
Thanks Chris!!

ckormanyos:
Thank you as well, John.
As a next course, it seems like the next priority is to take a look at `gmp_float` and `mpfr_float`.

PR #450, opened 2022-04-25
#include <iomanip> so that setprecision works.

NAThompson:


ckormanyos:
Hi Nick (@NAThompson) I also found this within the context of handling higher warning levels in `gmp.hpp` yesterday.

This change (including `<iomanip>` is also included in #448 which is cycling now.

Another header was (is) also missing on some compilation systems, namely `<cstring>`, since functions like `std::strlen()` are used in `gmp.hpp` as well.

Both of these headers are now included in my other PR. How would you like to handle this?

ckormanyos:
I suppose there is a chance that this code phenomenon might be reported in this trimester? I was thinking that if any clients need to work around this during the time of 1.79, they might have to include those headers above `gmp.hpp` in the client code?

NAThompson:
@ckormanyos : I don't have a preferred timescale for this-I'd let your stuff ripen and do a patch release then.

Issue #451, opened 2022-04-26
GMP backend fails to catch some zero denominators


ckormanyos:
In the GMP backend in `gmp.hpp`, some uses of quotients such as in [here](https://github.com/boostorg/multiprecision/blob/5c773e67703d9f0a896209de47f00f9253c3a3ca/include/boost/multiprecision/gmp.hpp#L3004) have a _hidden_ quotient and the code does not explicitly check for divide by zero.

ckormanyos:
See also the discussion [here](https://github.com/boostorg/multiprecision/pull/448#issuecomment-1108425173)

ckormanyos:
This issue is being handled in #452

ckormanyos:
Fixed by #452 and with larger scope in `number` class design by #453

PR #452, opened 2022-04-26
Catch divide by zero in GMP rational assignments

ckormanyos:
This is intended to handle #451

Hi John (@jzmaddock) here is a first cut attemting to catch div-by-zero in GMP rational assignments/ctors.

- Could you please take a look if I found all (and the right) places?
- Also I was a bit concerned for handling the move assignments properly, and I'm unsure if you agree with what i did?
- I wonder if we should add some explicit GMP-only tests in the arithmetic-tests? Such that these try and catch div by zero?

I also handled that single remaining MSVC Level 4 warning here (although it's a bit off topic).

### DIFF
--- a/include/boost/multiprecision/gmp.hpp
+++ b/include/boost/multiprecision/gmp.hpp
@@ -1753,6 +1753,10 @@ inline typename std::enable_if<boost::multiprecision::detail::is_arithmetic<T>::
    return a.compare(b) > 0;
 }
 
+inline bool eval_is_zero_move(gmp_int&& val)
+{
+   return mpz_sgn(val.data()) == 0;
+}
 inline bool eval_is_zero(const gmp_int& val)
 {
    return mpz_sgn(val.data()) == 0;
@@ -2380,6 +2384,9 @@ struct gmp_rational
    template <class T, class U>
    gmp_rational(const T& a, const U& b, typename std::enable_if<std::is_constructible<gmp_int, T>::value && std::is_constructible<gmp_int, U>::value>::type* = nullptr)
    {
+      if (b == static_cast<U>(0))
+         BOOST_MP_THROW_EXCEPTION(std::overflow_error("Division by zero."));
+
       gmp_int i(a), j(b);
       m_data[0]._mp_num = i.data()[0];
       m_data[0]._mp_den = j.data()[0];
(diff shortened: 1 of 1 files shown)


include/boost/multiprecision/gmp.hpp
```diff
@@ -2380,6 +2384,9 @@ struct gmp_rational
    template <class T, class U>
    gmp_rational(const T& a, const U& b, typename std::enable_if<std::is_constructible<gmp_int, T>::value && std::is_constructible<gmp_int, U>::value>::type* = nullptr)
    {
+      if (b == static_cast<U>(0))
```
jzmaddock, line 2387: I don't think we have tests that would catch this, but I would worry about T and U being types which gmp_int is constructible from, but which do not have an `operator=(int)`, an example would be `mpz_t`.

So I would be inclined to put this check after the construction of the gmp_int's below and use a call to eval_is_zero for the check.  Hope that makes sense!

include/boost/multiprecision/gmp.hpp
```diff
@@ -2390,6 +2397,9 @@ struct gmp_rational
    template <class U>
    gmp_rational(const gmp_int& a, const U& b, typename std::enable_if<std::is_constructible<gmp_int, U>::value>::type* = nullptr)
    {
+      if (b == static_cast<U>(0))
```
jzmaddock, line 2400: same here, and in a few places below.

jzmaddock:
There's a couple of comments attached to source, but I'll work up some generic tests to go in test_arithmetic.hpp and then we can see what breaks ;)

include/boost/multiprecision/gmp.hpp
```diff
@@ -1753,6 +1753,10 @@ inline typename std::enable_if<boost::multiprecision::detail::is_arithmetic<T>::
    return a.compare(b) > 0;
 }
 
+inline bool eval_is_zero_move(gmp_int&& val)
```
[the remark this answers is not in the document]
ckormanyos: Agreed. Thanks John. I was unsure of that conversion.

include/boost/multiprecision/gmp.hpp
ckormanyos: I understand. This makes sense. i will tweak these and similar ones accordingly.

jzmaddock:
Tests are coming.... there are more failures with cpp_rational/rational_adaptor than I was expecting :(

ckormanyos:
> Tests are coming.... there are more failures with cpp_rational/rational_adaptor than I was expecting

Understood.

Sorry John, I think I might have moved too quickly forward on merging. I believe we will need a new PR for adding the tests...

I will wait on any/all further activities until we resolve this phenomenon.

PR #457, opened 2022-05-04
Try handle about 60-mpfr and 1-bin-float warns

ckormanyos:


include/boost/multiprecision/cpp_bin_float/io.hpp
```diff
@@ -531,7 +531,12 @@ std::string cpp_bin_float<Digits, DigitBase, Allocator, Exponent, MinE, MaxE>::s
 #else
       std::intmax_t max_bits = cpp_bin_float<Digits, DigitBase, Allocator, Exponent, MinE, MaxE>::bit_count + ((cpp_bin_float<Digits, DigitBase, Allocator, Exponent, MinE, MaxE>::bit_count % limb_bits) ? (limb_bits - cpp_bin_float<Digits, DigitBase, Allocator, Exponent, MinE, MaxE>::bit_count % limb_bits) : 0) + limb_bits;
       if (power10)
-         max_bits += (msb(boost::multiprecision::detail::abs(power10)) / 8) * limb_bits;
+      {
```
jzmaddock, line 534: Since max_bits can never be negative, would it simplify things to make it unsigned?  Or does that complicate things elsewhere?

ckormanyos: Good point, John. I'll take a look when dealing with the next batch of MPFR warnings.
This one showed up in bin-float, but a little rework will be no problem if it seems easier.
I'll check it out.

ckormanyos: Actually when I tried making `max_bits` to be unsigned, there were some errors and warnings like here in my [prototyping project CI](https://github.com/ckormanyos/boost_multiprecision_harden/actions/runs/2272093129).

So I decided to leave it signed. I guess we could follow up later, but as it was previously signed, i thought it's OK.

Issue #429, opened 2022-02-02
Take a glance at GCC 4(.8) in CI


ckormanyos:
Take a glance at GCC 4(.8) in CI and see how it goes. If it works well consider dealing with it as was done for GCC 5 in #425.

ckormanyos:
This issue is phrased as a question. Does GCC 4(.8) run out-of-the-box on some CI tests when spun up in an (older LTS) Ubuntu CI machine? If so, consider adding such tests.

ckormanyos:
Here is my report on taking a glance at GCC 4.8 running on some CI tests.

I found the following relevant issues.

1. GCC 4.8 on Bionic does not recognize [this](https://github.com/boostorg/multiprecision/blob/ef6ad5e40f60cd3783de325031e8ce4a3ae6935d/include/boost/multiprecision/detail/number_base.hpp#L48) compiler check. I did [this](https://github.com/boostorg/multiprecision/blob/d20ee4079cd0f6c00d405c27822561262e16b80e/include/boost/multiprecision/detail/number_base.hpp#L48) potential workaround.
2. GCC 4.8 on bionic does not auto-detect `BOOST_MATH_STANDALONE` ( in Math, not Multiprecision) as the default. So I need to specify that definition either on the command line or above the `#include`s.
3. Some places in our code have mismatch `static constexpr` versus inline definition of const (in classes). Like [this here](https://github.com/boostorg/multiprecision/blob/ef6ad5e40f60cd3783de325031e8ce4a3ae6935d/include/boost/multiprecision/cpp_int.hpp#L525) where the calss uses `static constexpr` but the code uses `const`. So I needed to make sure the word `constexpr` agrees in the class as well as the code. This happened about ten times. I corrected in my branch. There are a few in Math, but that`'s another topic. GCC 4.8 found these qhereas others did not.
4. I can not get anything actually to run on GCC 4.8 in CI. I reduced CI to the candidate 4.8 job only, but I cant see any codes running.

I would actually like help of point 4, since I would like to see something running. My separate projects do just fine with the above-mentioned changes in regard to running with Multiprecision together on 4.8.

Now I had an agreement with Matt (@mborland) that 4.8 needs to run out-of-the-box, or I'll give up. i'm fine with giving up, but this thing is really close.

Thoughts? Comments?

Cc: @jzmaddock and @mborland and @NAThompson

NAThompson:
[...]

PR #458, opened 2022-05-07
Try (with minimal effort) GCC4 in arithmetic tests

ckormanyos:


ckormanyos:
The result of this PR so far is:

- Do not include GCC 4.8 runs in CI.
- Make a separate file `detail/check_cpp11_config.hpp` capable of local specialization if needed.
- Repair numerous instances of mismatch `const`-versus-`constexpr`.
- Repair consistency in all header guards according to use `BOOST_MP_`-plus-filename-`_HPP` and optionally add uniqueness (e.g. `_DETAIL`) as needed.

ckormanyos:
Hi John (@jzmaddock) this thing is review-ready. I'll struggle no further on it until we reach a 4.8 consensus. At the moment there is no effort to fully incorporate 4.8.

This PR in this state officially closes #429. You can, however, also choose to pursue this thread further. I'm OK with that too. Or later get back on it, or whatever?

I do, however, really want the changes within and as sketched/described above. Also a few changes in Math needed, as these are bugs anyway, I'll PR them shortly.

Tell me what you think...? And we can move on.

Cc: @mborland and @NAThompson

ckormanyos:
Any feedback on this? It seems to be running green in CI.

jzmaddock:
Sorry for the delay, yep, go for it!

ckormanyos:
>  yep, go for it!

Thanks John. This PR ended up to not go entirely for GCC 4.8 and there is no 48. added run in CI. Multiprecision was pretty close to accepting 4.8, but a few small issues prevented that.

GCC 4.8 did find some inconsistencies (mostly `const`-versus-`constexpr`) as mentioned above and this PR resolves these nicely.

Cc: @mborland and @NAThompson

PR #461, opened 2022-05-17
add missing include of memory

svigerske:
With GCC 12, header memory is included by less other system headers: https://gcc.gnu.org/gcc-12/porting_to.html

However, it is required in float128.hpp due to the use of `std::unique_ptr`.


ckormanyos:
See also #462

ckormanyos:
Thanks for this catch @svigerske.

I decided to simply merge this one, then let CI run. After that, I'll add GCC12/clang++ on Jammy in a separate PR. Well, maybe this wasn't the easiest path (from my side) for this combined effort, but slow and steady it'll work.

PR #463, opened 2022-05-19
Closer look gcc12 clang14 jammy

ckormanyos:


ckormanyos:
Hi John (@jzmaddock) the initial tests on GCC12/clang14 in Jammy were mostly good. There is, however, one fail on clang14 using C++2b. It actually appears to be a legitimat error and not a timeout.

I have not yet taken a closer look, but it seems like there might be a legit issue needing attention (if, that is, we'd like to compile with experimental `-std=c++2b`).

jzmaddock:
Looks like an issue to me, will probably need a bit of "comparative debugging" to see why gcc and clang are giving different answers.  I don't have these compilers installed locally yet... still on the todo list!

ckormanyos:
>  I don't have these compilers installed locally yet... still on the todo list!

Yeah, me too. I've been _getting_ _lucky_ on the server farms.

So... how would you like to proceed? It seems like the defect(s) are/were only found on one compiler with `-std=c++2b`.
- I/we could eliminate that language standard from the test matrix and integrate the CI run, and add that standard back to the CI run(s) later.
- Or we could wait a while and see if the problem can be isolated, then add the entire CI run(s)?

jzmaddock:
I would suspect a compiler or std lib issue, but you never know.

I'm downloading Ubuntu-22 now, so give me a few days and I *may* have an answer...

ckormanyos:
> suspect a compiler or std lib issue, but you never know.

That makes sense. Indeed, I'm also sort of _feeling_ a compiler issue here.

> I'm downloading Ubuntu-22 now, so give me a few days and I _may_ have an answer...

Thanks John... Seems like it might be a tough one to narrow down.

jzmaddock:
One fix pushed, there is still a further issue that `std::is_constant_evaluated()` is returning true when it should not.  I need to investigate that further...

ckormanyos:
Just peered at this... Maybe a `break` instead of the comment ` // fallthrough`, but I'm not entirely sure on the intent...

jzmaddock:
Grrr, uploading legal code would have helped!  Hopefully corrected now.

jzmaddock:
The clang issue is reported here: https://github.com/llvm/llvm-project/issues/55638

We need to work around this as well as otherwise we call sub-optimal routines as it assumes everything is constexpr all the time!

Issue #464, opened 2022-05-28
cpp_dec_float<>::convert_to<double>() is locale dependent


StefanBruens:
When a locale is used which does not use '.' as decimal point, `convert_to<double>` fails:

```
#include <boost/multiprecision/cpp_dec_float.hpp>
#include <iostream>
#include <locale>

int main (int argc, char* argv[])
{
  auto a = boost::multiprecision::cpp_dec_float_50{12345};

  std::cout << std::scientific << std::setprecision(8);
  double d1 = a.convert_to<double>();
  std::cout << "d1 = " << d1 << std::endl;

  std::locale::global(std::locale("de_DE"));
  double d2 = a.convert_to<double>();
  std::cout << "d2 = " << d2 << std::endl;

  return 0;
}
```
```
g++ -o boost_mp boost_mp.cpp -g3 && ./boost_mp 
d1 = 1.23450000e+04
d2 = 1.00000000e+00
```

This is probably a regression due to:
https://github.com/boostorg/multiprecision/commit/93d7e83d72e33e19183ff51fd65d56ba5c70860d

The intermediate string always uses a '.' as decimal point:
https://github.com/boostorg/multiprecision/blob/37f1c76375915493cf43621b333549671809071b/include/boost/multiprecision/detail/number_base.hpp#L1479

but it is parsed in a locale dependent way, with std::strtod:
https://github.com/boostorg/multiprecision/blob/37f1c76375915493cf43621b333549671809071b/include/boost/multiprecision/cpp_dec_float.hpp#L1606

mborland:
@jzmaddock and @ckormanyos I am pretty sure I broke this removing Boost.lexical_cast and string streaming. Investigating.

ckormanyos:
Thanks for looking into this Matt. I'm not entirely sure, but unfortunately, it looks like GCC5 on bionic is having [difficulty](https://github.com/boostorg/multiprecision/runs/6637778256?check_suite_focus=true#step:17:433) with locales.

No problem from my side to disable those tests/functionalities locally for `(#defined (__GNUC__) && (__GNUC__ < 6))`. I don't know if this helps?

StefanBruens:
> Thanks for looking into this Matt. I'm not entirely sure, but unfortunately, it looks like GCC5 on bionic is having [difficulty](https://github.com/boostorg/multiprecision/runs/6637778256?check_suite_focus=true#step:17:433) with locales.
> 
[...]

PR #465, opened 2022-05-28
Fix for issue #464

mborland:
Reverts locale dependent string conversion with string streaming


mborland:
@ckormanyos this should be good to go based on feedback from @StefanBruens.

ckormanyos:
Thank you for reporting this Stefan (@StefanBruens).

> good to go

Cool. Thank you for fixing this, Matt (@mborland). Merged.

PR #470, opened 2022-06-05
Enhanced gcc clang warnings

ckormanyos:
This PR attempts to handle the remaining conversion warnings on the MPFR backend. This will wrap up the effort to handle enhanced GCC/clang/MSVC warnings in this trimester, as the other backends are finished.

Some remaining work is probably lurking around, undiscovered, in various less-well-used parts of Multiprecision. The entire warning-hardening drive, however, has resulted in hundreds of edits and running tests.

Let's see how this goes in CI...?

### DIFF
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -382,11 +382,18 @@ class cpp_dec_float
    {
       using default_ops::eval_add;
       using default_ops::eval_multiply;
-      static constexpr unsigned          bit_shift = sizeof(unsigned long long) * CHAR_BIT;
+
+      static constexpr unsigned     bit_shift = sizeof(unsigned long long) * CHAR_BIT;
       static constexpr uint128_type mask      = (static_cast<uint128_type>(1u) << bit_shift) - 1;
-      unsigned                           shift     = bit_shift;
-      *this                                        = static_cast<unsigned long long>(v & mask);
+
+      unsigned shift = bit_shift;
+
+      static_cast<void>(shift);
+
+      *this = static_cast<unsigned long long>(v & mask);
+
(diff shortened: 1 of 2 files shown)


include/boost/multiprecision/cpp_dec_float.hpp
```diff
@@ -382,11 +382,18 @@ class cpp_dec_float
    {
       using default_ops::eval_add;
       using default_ops::eval_multiply;
-      static constexpr unsigned          bit_shift = sizeof(unsigned long long) * CHAR_BIT;
+
+      static constexpr unsigned     bit_shift = sizeof(unsigned long long) * CHAR_BIT;
       static constexpr uint128_type mask      = (static_cast<uint128_type>(1u) << bit_shift) - 1;
-      unsigned                           shift     = bit_shift;
```
jzmaddock, line 387: That must have been a cut-and-paste error, variable shift is unused and can be deleted rather than cast to void.

Other than that, this all looks great to me!

ckormanyos:
> Other than that, this all looks great to me!

I has thought so also, ... until I, oops, used too much of 14...

Oops. I got so pumped up on C++14, that I ended up using a couple of instantiations of `std::make_unsigned_t<>`. Of course this is not yet allowed since we still test on C++11.

So I will need to remove these tomorrow.

include/boost/multiprecision/cpp_dec_float.hpp
ckormanyos: I forgot this in my recent push to green. I am going to commit and merge to develop anyway and cycle back to this trivial removal later.

ckormanyos:
> ended up using a couple of instantiations of `std::make_unsigned_t<>`

> ... will need to remove these tomorrow.

Done. Green

Issue #475, opened 2022-06-14
Intel intrinsics erroneously included when compiling on Windows with clang-cl for ARM64


tailsu:
The current MSVC workaround [here](https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/cpp_int/intel_intrinsics.hpp#L38-L44) is insufficient to handle the case.

When compiling with clang on Windows, the `immintrin.h` header is available when compiling for ARM64, but `BOOST_MSVC` is not defined, so `BOOST_MP_HAS_IMMINTRIN_H` doesn't get `#undef'd` and compilation fails down the road.

You can reproduce the issue by compiling some demo program with the `clang_cl_arm64_x64` toolset in Visual Studio. You'll need to install clang and clang-cl with the Visual Studio Installer first, of course.

jzmaddock:
Fix merged, closing.

PR #476, opened 2022-06-14
explicitly check for ARM/ARM64 on Windows

tailsu:
Tentative fix for https://github.com/boostorg/multiprecision/issues/475

### DIFF
--- a/include/boost/multiprecision/cpp_int/intel_intrinsics.hpp
+++ b/include/boost/multiprecision/cpp_int/intel_intrinsics.hpp
@@ -35,10 +35,12 @@
 #undef BOOST_MP_HAS_IMMINTRIN_H
 #endif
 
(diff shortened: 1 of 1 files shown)


ckormanyos:
Thanks for looking into this @tailsu.

I have approved the CI run (for first-time contributors).

There was some similar discussion a few months ago, si we need to ask @jzmaddock if this is a similar case and if the fix is OK.

tailsu:
OK, I'm glad 👍 

It would also make sense to add the compiler/platform combo to CI but I'm not sure how to do that.

mborland:
@jzmaddock and @ckormanyos this should probably make the release as well since it is linked to multiple issues now.

ckormanyos:
Thanks @tailsu, John and Matt for driving this forward.

PR #477, opened 2022-06-23
number: fix multiplication and division when precision differs

mniestroj:
This is clearly a copy-paste bug, as there is addition in both places
now. Fix that.

Fixes: 26ab08df2ed0 ("Test and fix mixed precision arithmetic with
  component types.")


jzmaddock:
Ugh, that's horrible, thanks for catching this!

ckormanyos:
Nice catch! Thank you.

jzmaddock:
Failing tests are caused by an unrelated change from gcc-8 mingw to gcc-11.2 mingw.  Strangely I'm unable to reproduce locally, and it looks a lot like a broken install (or binary path?) on the testing image, but it's very hard to tell.  In any case, I'm going to go ahead and merge while we figure this out.

We should also try to add some tests for this issue.

PR #478, opened 2022-06-24
probe issue with gcc on windows.

jzmaddock:


ckormanyos:
Thank you John (@jzmaddock)! Just in time for 1.80. Wow!

Issue #485, opened 2022-08-20
New 1.80 Warnings with BOOST_NO_EXCEPTIONS in `cpp_int.hpp`


ckormanyos:
MSVC, clang and some GCC versions detect warnings when using `BOOST_NO_EXCEPTIONS` around here at lines [1810](https://github.com/boostorg/multiprecision/blob/a4cd4ffe2f4c008e0fb2cdeeae1ff2a8c62efe51/include/boost/multiprecision/cpp_int.hpp#L1810), [1846](https://github.com/boostorg/multiprecision/blob/a4cd4ffe2f4c008e0fb2cdeeae1ff2a8c62efe51/include/boost/multiprecision/cpp_int.hpp#L1846) and [1884](https://github.com/boostorg/multiprecision/blob/a4cd4ffe2f4c008e0fb2cdeeae1ff2a8c62efe51/include/boost/multiprecision/cpp_int.hpp#L1884).

I'm a bit surprised I missed these while ramping up for 1.80. Sorry I missed them at that time.

There are a few ways to fix these. Either initialize the variable `val` at its inception, or initialize it within the places where the exception stuff expands to nothing.

I would initialize `val` in the scopes where the exception stuff expands to nothing. Does anyone have a different preference of how to fix them?

Otherwise, I'll fix them and PR the changes ASAP.

Cc: @jzmaddock and @mborland

PR #486, opened 2022-08-20
Handle noinit warnins on BOOST_NO_EXCEPTIONS

ckormanyos:
This PR is intended to handle #485. We can negotiate on the best edits for the patch, but these should do the trick.

Cc: @jzmaddock and @mborland


ckormanyos:
I'll run a few more checks with some other backends on no-exceptions. For now, Ican't find any more, but maybe a few similar issues might pop up. I'll try to look around for any similar ones.

Anyway, I'll merge this for now, as it handles `cpp-int`.

Issue #499, opened 2022-10-05
cpp_dec_float incorrect constructing from string


yaroslavchahovets:
When constructing **cpp_dec_float** from certain string values which don't represent correct numbers, we expect that the exception is thrown, but the object is successfully constructed:

```
#include <iostream>
#include <boost/multiprecision/cpp_dec_float.hpp>

using boost::multiprecision::cpp_dec_float_50;

int main()
{
    std::cout.precision(10);

    cpp_dec_float_50 val1{"12a3.4"};
    std::cout << "no exception. val1=" << val1 << "\n";

    cpp_dec_float_50 val2{"1.2a34"};
    std::cout << "no exception. val2=" << val2 << "\n";

    return 0;
}

Output:
no exception. val1=12.4
no exception. val2=1.00000002
```
This bug is reproduced on Boost 1.79.0 and 1.80.0. Also it is reproduced on current repository https://github.com/boostorg/boost.git (as of 2022-10-05). The behaviour is correct on Boost 1.78.0. Compilers: gcc 9.4.0 and gcc 10.3.0.

ckormanyos:
> When constructing `cpp_dec_float` from certain string values which don't represent correct numbers, we expect that the exception is thrown, but the object is successfully constructed

OK. Thanks for reporting this, @yaroslavchahovets.

> bug is reproduced

I'm not sure about the term _bug_ in this case. We simply need to define the proper behavior.

> The behaviour is correct on Boost 1.78.0

What do you think would be the correct behavior? Throw an exceptuion seems to be agreed. What value would be expected. I might prefer NaN.

Thoughts?

Cc: @jzmaddock

yaroslavchahovets:
> What do you think would be the correct behavior? Throw an exceptuion seems to be agreed.

I think that correct behavior is to do what boost_multiprecision up to version 1_78_0 did in this case: throw **std::runtime_error** with explanatory string (what() return value) `Unable to parse the string "<string_argument>" as a floating point value.`
For the example program above: throw **std::runtime_error** with explanatory string `Unable to parse the string "12a3.4" as a floating point value.`

By the way, the code which do exactly this is in the source code in https://github.com/boostorg/multiprecision/blob/c6794e5638f9bf942eafe33da0f8940649696674/include/boost/multiprecision/cpp_dec_float.hpp#L2383-L2386

yaroslavchahovets:
[...]

PR #502, opened 2022-10-12
Fix for issue 499

mborland:
Replaced boost::lexical_cast with std::stol without updating the type of exception being caught

### DIFF
--- a/include/boost/multiprecision/cpp_dec_float.hpp
+++ b/include/boost/multiprecision/cpp_dec_float.hpp
@@ -2106,11 +2106,7 @@ bool cpp_dec_float<Digits10, ExponentType, Allocator>::rd_string(const char* con
       if (((pos = str.find('e')) != std::string::npos) || ((pos = str.find('E')) != std::string::npos))
       {
          // Remove the exponent part from the string.
-         #ifndef BOOST_MP_STANDALONE
-         exp = boost::lexical_cast<exponent_type>(static_cast<const char*>(str.c_str() + (pos + 1u)));
-         #else
          exp = static_cast<exponent_type>(std::atoll(static_cast<const char*>(str.c_str() + (pos + 1u))));
-         #endif
          
          str = str.substr(static_cast<std::size_t>(0u), pos);
       }
@@ -2373,11 +2369,7 @@ bool cpp_dec_float<Digits10, ExponentType, Allocator>::rd_string(const char* con
 
 #ifndef BOOST_NO_EXCEPTIONS
    }
-   #ifndef BOOST_MP_STANDALONE
-   catch (const bad_lexical_cast&)
-   #else
    catch (const std::exception&)
-   #endif
    {
       // Rethrow with better error message:
       std::string msg = "Unable to parse the string \"";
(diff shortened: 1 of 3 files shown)


jzmaddock:
Does this work?  strtoll is non-throwing.

mborland:
> Does this work? strtoll is non-throwing.

This uses `stol` which throws `std::invalid_argument` like the issue was seeing as opposed to returning 0 like `strtol`. The catch block was only for `bad_lexical_cast` which was causing `std::invalid_argument` to propagate to the user.

ckormanyos:
Thank you for going after this, Matt (@mborland)!

yaroslavchahovets:
Hello!
I checked this fix (commit 60a40c598873c47e8e5e94d438eab1bc913741b0) and found the following.
1. It fixes the case when input string is "wrong".
2. It does not fix cases when input string is "12a3.4" and "1.2a34" (see https://github.com/boostorg/multiprecision/issues/499#issue-1398168644)
3. It introduces some regress. When input string has scientific notation, for example "3.4e1a2", the expected result is throwing the std::runtime_exception with the text "Unable to parse the string \"3.4e1a2\" as a floating point value.", whereas with this fix, the test program generates no exception, result is 34.

So I think this commit (60a40c5988) does a good job, but another commit(s) should be added to fix all the cases.
Previously I proposed the idea for fix  "12a3.4" and "1.2a34" cases here: https://github.com/boostorg/multiprecision/issues/499#issuecomment-1270023503

Thank you for your work, @mborland .

mborland:
@yaroslavchahovets This next pair of commits should address your other cases. Thanks for the feedback.

yaroslavchahovets:
Now all cases I reported previously ("wrong", "12a3.4", "1.2a34", "3.4e1a2") produce expected behavior.
But the cases "12L3.4", "12f3.4" produce wrong behavior (no exceptions, value is 12.4).
Also the cases "1.2L34", "1.2f34" produce wrong behavior (no exceptions, value is 1.00000002).

Thank you.

yaroslavchahovets:
Hello. Commit https://github.com/boostorg/multiprecision/pull/502/commits/e1abb81da102d86046e8381bd78e27a67fad7b9a fixes behavior for cases "12L3.4", "12f3.4", thanks.
But for the 'normal' cases "1.2f", "12f", "12L" it throws an exception. I think the correct behavior for these cases is to produce values 1.2, 12, 12 respectively.

yaroslavchahovets:
Or, alternatively, we can consider format with F,f,L,l suffixes illegal (as it was in boost 1.78.0 and earlier). In this case we should keep the behavior of commit https://github.com/boostorg/multiprecision/pull/502/commits/e1abb81da102d86046e8381bd78e27a67fad7b9a  and delete all the code which implements functionality of F,f,L,l suffixes.

jzmaddock:
I would reasonably strongly favour keep suffixes as illegal: it's not clear what meaning they confer in a multiprecision context.

yaroslavchahovets:
@jzmaddock Please consider integrate this pull request into the main repository

jzmaddock:
Sorry for being late to review this, just checking this out, we still erroneously parse things like "23.5-3e-26"

mborland:
@jzmaddock since I am a novice at regex can you take a look at this? The expression is [here](https://regex101.com/r/D5H8xT/1) and should match any valid number. I am receiving 0 matches once plugged into `std::regex_search`.

jzmaddock:
Fixed in alternate PR, closing.

PR #522, opened 2023-01-17
Improve printing out configuration info in CI runs.

jzmaddock:
@ckormanyos This PR should (hopefully) improve the configuration info that's printed out with each CI run, including the b2 configuration logs.

### DIFF
--- a/.drone/boost.sh
+++ b/.drone/boost.sh
@@ -35,7 +35,7 @@ echo '==================================> BEFORE_SCRIPT'
 echo '==================================> SCRIPT'
 
 echo "using $TOOLSET : : $COMPILER : <cxxflags>-std=$CXXSTD ;" > ~/user-config.jam
-(cd libs/config/test && ../../../b2 config_info_travis_install toolset=$TOOLSET && ./config_info_travis)
+(cd libs/config/test && ../../../b2 print_config_info print_math_info toolset=$TOOLSET)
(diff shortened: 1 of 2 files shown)


ckormanyos:
> PR should (hopefully) improve the configuration info that's printed out with each CI run

Thanks John! That's going to be great when the kinks get ironed out.

If we like it and it's stable, we could consider `grep`-ing, `gawk`-ing or `sed`-ing, for even more summaries. I could look into that if we decide there might be (even more) added value to gain...

jzmaddock:
Hopefully small kinks, some useful info though like:

```
gcc.compile.c++ ../../../bin.v2/libs/multiprecision/config/gcc-10/debug/threading-multi/visibility-hidden/has_gmp.o
In file included from ../../../boost/config/detail/suffix.hpp:572,
                 from ../../../boost/config.hpp:61,
                 from ../config/has_gmp.cpp:8:
../config/has_gmp.cpp:11:69: note: ‘#pragma message: __GNU_MP_VERSION=6’
   11 | #pragma message "__GNU_MP_VERSION=" BOOST_STRINGIZE(__GNU_MP_VERSION)
      |                                                                     ^
../config/has_gmp.cpp:12:81: note: ‘#pragma message: __GNU_MP_VERSION_MINOR=2’
   12 | #pragma message "__GNU_MP_VERSION_MINOR=" BOOST_STRINGIZE(__GNU_MP_VERSION_MINOR)
      |                                                                                 ^
```

That could probably be made neater by printing it all out from a config_info like program, but it'll do for now.

jzmaddock:
Failure is spurious, merging.

Issue #526, opened 2023-02-05
Loading cpp_int by std::string branches on uninitialized memory


guidovranken:
Program:

```cpp
#include <boost/multiprecision/cpp_int.hpp>
int main(void)
{
    using namespace boost::multiprecision;
    const std::string v("00000000000000000000000052222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222");
    cpp_int V(v);
    return 0;
}
```

Compile:

```sh
$ clang++ -I boost_1_81_0/ p.cpp -g
```

Run under valgrind:
```
$ valgrind ./a.out
==2920517== Command: ./a.out
==2920517== 
==2920517== Conditional jump or move depends on uninitialised value(s)
==2920517==    at 0x4022B7: boost::multiprecision::backends::cpp_int_base<0ul, 18446744073709551615ul, (boost::multiprecision::cpp_integer_type)1, (boost::multiprecision::cpp_int_check_type)0, std::allocator<unsigned long long>, false>::normalize() (cpp_int.hpp:405)
==2920517==    by 0x401CEA: boost::multiprecision::backends::cpp_int_backend<0ul, 0ul, (boost::multiprecision::cpp_integer_type)1, (boost::multiprecision::cpp_int_check_type)0, std::allocator<unsigned long long> >::do_assign_string(char const*, std::integral_constant<bool, false> const&) (cpp_int.hpp:1874)
[...]

PR #527, opened 2023-02-05
Fix potentially uninitialized storage when parsing octal strings.

jzmaddock:
Fixes https://github.com/boostorg/multiprecision/issues/526


jzmaddock:
One failure looks spurious, merging.

PR #530, opened 2023-02-09
Fix redundant redeclaration when build with C++ >= 17.

Romain-Geissler-1A:
This fixes the following error when using C++ >= 17 and gcc 11:

```
/data/mwrep/res/osp/Boost/21-0-0-9/include/boost/multiprecision/cpp_int/literals.hpp:220:13: error: redundant redeclaration of 'constexpr' static data member 'boost::multiprecision::literals::detail::make_backend_from_pack<Pack, B>::value' [-Werror=deprecated]
```

### DIFF
--- a/include/boost/multiprecision/cpp_int/literals.hpp
+++ b/include/boost/multiprecision/cpp_int/literals.hpp
@@ -212,8 +212,10 @@ struct make_backend_from_pack
    static constexpr B value = p;
 };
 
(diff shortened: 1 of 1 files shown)


mborland (approved):

mborland:
Looks good to me. This was deprecated in C++17 as they are now implicitly inlined.

Issue #297, opened 2021-02-11
Binary string to cpp_int


BoredBored:
Hello, I noticed boost multiprecision does not have a binary string to decimal converter, so I made one. Hope you enjoy!
```c++
// C++ program to convert binary to decimal
// C++ program to convert binary to decimal
#include <iostream>
#include <boost/multiprecision/cpp_int.hpp>

boost::multiprecision::cpp_int binaryToDecimal(const std::string_view& num)
{

    boost::multiprecision::cpp_int dec_value = 0;
    auto c = num.data();
    auto len = num.size();
    //check if big enough to have 0b postfix
    if(num.size() > 2){
        //check if 2nd character is the binary postfix
        //skip over it & adjust length accordingly if it does
        if(num[1] == 'b' || num[1] == 'B'){
            c += 2;
            len -= 2;
        }
    }
    /*
     *  Read digits left to right
     *  Ex.
     *   (0b postfix is ignored)
     *
     *               { '1', '1', '1', '0' }
     *   str index:     0    1    2    3
     *   bin index      3    2    1    0
     *
     *   The bit index is inverse (displayed above) to the
     *   digits in the string, so we must the algorithm will be
     *
     *   cpp_int dec_value = 0;
     *   int     len       = num.size();
     *   int     bin_index = len - 1;
     *   int     str_index = 0;
     *
     *   for(; str_index < len; ++str_index, --bin_index){
     *      dec_value |= (str[str_index] - '0') << bin_index;
     *   }
     *   return dec_value;
     */
    for (int_fast16_t i = len - 1; 0 <= i; ++c, --i) {
        dec_value |= (*c - '0') << i;
    }
    return dec_value;
}

// Driver program to test above function
int main()
{
    std::cout << binaryToDecimal("0b000000111") << '\n';
}
```

ckormanyos:
This is an interesting post. Boost.Multiprecision is transitioning to C++11 for the next release. Literal binary strings, if memory serves, arrived in C++14. Boost.Multiprecision is, however, not strictly tied to that exact language evolutionary step.

Does auto-detect of binary-string-to-multiprecision type make sense? And if so, for which (or all) or our backends?
Any interest in digit separators, or is this a separate issue altogether?

jzmaddock:
There are really two features here:

* iostream support for binary strings (which isn't in C++14 in fact, but no matter).
* Literal/constexpr support via user defined literals.

[...]

PR #535, opened 2023-02-19
Add support for std::format.

jzmaddock:
This is a tentative first step towards supporting C++20's std::format.

Integers are all working and feature complete other than the usual undiscovered bugs and corner cases.

Other number types will require some refactoring of the code in format.hpp, though exactly how/what remains to be seen.

This also ties in with https://github.com/boostorg/multiprecision/issues/297 as it adds binary string formatting.

Currently only latest msvc supports `<format>`, for gcc we will have to wait for gcc-13 later in the year, likewise clang/libc++.

Comments welcome as always.

### DIFF
--- a/include/boost/multiprecision/detail/format.hpp
+++ b/include/boost/multiprecision/detail/format.hpp
@@ -1,0 +1,365 @@
+///////////////////////////////////////////////////////////////////////////////
+//  Copyright 2023 John Maddock. Distributed under the Boost
+//  Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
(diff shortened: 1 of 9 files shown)


DouisLavid:
Hello,
Thanks for the awesome work in Boost.Math :)
I was wondering if there was any plan to merge this PR in the foreseeable future, as it is a feature that would be quite useful for me. If not, would you say that there is any reason that would make me *not* want to use your changes locally ? (I'm ok with the "usual undiscovered bugs and corner cases" as nothing display related is critical in my use case, and i'm ok with breaking changes when the feature gets officially supported in the future)

jzmaddock:
Please do go ahead and use this and report back: we need to revisit this now that latest GCC supports std::format and finish this support off.

PR #539, opened 2023-03-07
Reduce testing resource usage

mborland:
See: https://github.com/boostorg/math/issues/964

### DIFF
--- a/.github/workflows/multiprecision.yml
+++ b/.github/workflows/multiprecision.yml
@@ -77,7 +77,7 @@ jobs:
         run: ../../../b2 print_config_info print_math_info toolset=$TOOLSET
         working-directory: ../boost-root/libs/config/test
       - name: Test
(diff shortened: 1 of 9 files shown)


mborland:
This dropped the total computational time in CI by ~15%. @jzmaddock is there anything else you can think of off hand that might be worth conditionally disabling to save resources?

jzmaddock:
> This dropped the total computational time in CI by ~15%. @jzmaddock is there anything else you can think of off hand that might be worth conditionally disabling to save resources?

Not off hand.  But we should probably have one MSVC and one clang runner running the MP tests too?

mborland:
All three are covered with the latest commit.

jzmaddock:
> All three are covered with the latest commit.

Thanks!

mborland:
Drone failure is on the CI system's end.

Issue #541, opened 2023-03-10
Problem with expression template


afabri:
Hello,

The code below does not compile with clang (but with VC2022) and boost 1_80_0.
The arithmetic expression `2 * r` becomes an `expression<..>` and that does not convert to the rational type.

Is that normal behavior, or a bug.

``` cpp
#include <boost/multiprecision/cpp_int.hpp>

typedef boost::multiprecision::cpp_rational Rat;

struct E {
  Rat rat;
  E(const Rat& rat)
    : rat(rat)
  {}
};

int main()
{
  Rat r(1,3);
  E e = 2 * r;
  // E e(2*r);           //compiles
 //  E e = Rat(2*r); // compiles 
  return 0;
}
```
https://gist.github.com/afabri/fcd036af9a1258b3d9d0bfd9977fd932

jzmaddock:
I believe this is expected behaviour - the issue being that there are **two** user-defined conversions required to construct an E from an expression template - one the conversion to Rat, and one in E's constructor.  This is certainly unfortunate, but I don't see an easy fix at our end right now.

jzmaddock:
@mborland : this is unrelated to the changes in rational_adaptor, I can reproduce with mpfr_float as well, it's intrinsic to the use of expression templates.

jzmaddock:
I might have a fix.

jzmaddock:
This is somewhat fixed in develop: the fix requires C++17 or later, prior to that the fix causes all sorts of issues with gcc and clang that I've been unable to fix.

jzmaddock:
This fix has been reverted as causing too many issues.

jzmaddock:
I'm going to close this as unfixable, consider:

```
#include <boost/multiprecision/cpp_int.hpp>
#include <cmath>

template <class Num>
struct wrapper
{
   wrapper(const Num&);
};

using N = boost::multiprecision::cpp_int;
using W = wrapper<N>;

bool foo(const N&);
bool foo(const W&);

int main(int argc)
{

   N n;
   W w(n);

   foo(n * n);

   return 0;
}
```

If we allow expression templates to convert to all the types that class `number` converts to, then this example will necessarily fail, as the call to `foo` will be ambiguous: there are two paths the compiler could take, each involving one user-defined conversion.

[...]

PR #543, opened 2023-03-11
Make expression templates implicitly convert...

jzmaddock:
to types that the number type will also convert to. Fixes: https://github.com/boostorg/multiprecision/issues/541.

### DIFF
--- a/include/boost/multiprecision/detail/number_base.hpp
+++ b/include/boost/multiprecision/detail/number_base.hpp
@@ -600,10 +600,20 @@ struct expression<tag, Arg1, void, void, void>
    template <class T
 #ifndef __SUNPRO_CC
              ,
-             typename std::enable_if<!is_number<T>::value && !std::is_convertible<result_type, T const&>::value && std::is_constructible<T, result_type>::value, int>::type = 0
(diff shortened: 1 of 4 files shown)


afabri:
Is this already in boost 1.82 ?

jzmaddock:
It is, however, this https://github.com/boostorg/multiprecision/pull/545 has already cropped up, and as I fear this is the tip of an iceberg shaped tricky problem, I'm contemplating pulling the fix from 1.82.

afabri:
It works for `mpq_class` which (we should double check) also uses expression templates.

lrineau:
I agree this PR should be reverted. The class `mpq_class` from `<gmpxx.h>` does not has those `operator T()` to convert expression to everything. Instead their equivalent of `number` has non-explicit conversion constructors from the expression templates.

Since this PR, Boost MP expression templates are convertible to too many types, and that can lead to compilation errors with ambiguous calls (like https://github.com/boostorg/multiprecision/pull/543#issuecomment-1481575350).

jzmaddock:
Reverted in develop and master.

Issue #544, opened 2023-03-17
Boost.Math multiprc_concept_check_6 failing with Clang-14 and C++17


mborland:
Under Clang-14 with C++17 there is a new error here:  https://github.com/boostorg/math/actions/runs/4447933117/jobs/7810075798#step:17:1835 saying that number base is an incomplete type. It doesn't look like any other version of clang is effected https://drone.cpp.al/boostorg/math/1267.

PR #545, opened 2023-03-20
Define BOOST_MP_NO_ET_IMPLICIT_CONVERSIONS for Clang 14 w/C++17 on Linux

mborland:
Workaround for #544

### DIFF
--- a/include/boost/multiprecision/detail/number_base.hpp
+++ b/include/boost/multiprecision/detail/number_base.hpp
@@ -113,6 +113,13 @@
 #  define BOOST_MP_NO_ET_IMPLICIT_CONVERSIONS
 #endif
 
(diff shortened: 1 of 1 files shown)


jzmaddock:
If this is required all the way up to clang-14, then the whole concept of this fix is probably not viable after all :(

But.... didn't CI with clang-14 pass without this define?  What's the test case - I wonder if the issue is dependent on the std lib is_constructible internals which would be horrible/worrying?

mborland:
It is only clang-14 that failed here: https://github.com/boostorg/math/actions/runs/4474375135/jobs/7862781129#step:17:315. Based on the error it looks like Clang is using the version of libstdc++ that comes with g++12, but g++12 compiles it fine.

PR #562, opened 2023-08-18
`denorm_min` is not zero

ryanelandt:
A type `T` is specialized if `std::numeric_limits<T>::is_specialized` is true. For specialized floating point types that support denorms, `denorm_min` should be less than `min`. For specialized floating point types that don't support denorms, `denorm_min` should be equal to `min`, e.g., see [this](https://en.cppreference.com/w/cpp/types/numeric_limits/denorm_min) or [this](https://github.com/boostorg/math/pull/1012#discussion_r1287102732).

The types `cpp_bin_float` and `cpp_dec_float` are floating point types that do not support denorms. This PR changes the values of `denorm_min` for these types from `0` to `min`.

### DIFF
--- a/test/test_cpp_bin_float.cpp
+++ b/test/test_cpp_bin_float.cpp
@@ -50,25 +50,29 @@ typedef number<cpp_bin_float<std::numeric_limits<good_type>::digits, digit_base_
 void test_special_cases()
 {
 #if !defined(BOOST_CI_ASAN_BUILD) && !defined(BOOST_CI_USAN_BUID)
-   test_type max_val     = (std::numeric_limits<test_type>::max)();
-   test_type min_val     = (std::numeric_limits<test_type>::min)();
-   test_type eps         = std::numeric_limits<test_type>::epsilon();
-   test_type inf_val     = (std::numeric_limits<test_type>::infinity)();
-   test_type nan_val     = (std::numeric_limits<test_type>::quiet_NaN)();
-   test_type half        = 0.5;
-   test_type one_point_5 = 1.5;
+   test_type max_val        = (std::numeric_limits<test_type>::max)();
+   test_type denorm_min_val = std::numeric_limits<test_type>::denorm_min();
+   test_type min_val        = (std::numeric_limits<test_type>::min)();
+   test_type eps            = std::numeric_limits<test_type>::epsilon();
+   test_type inf_val        = (std::numeric_limits<test_type>::infinity)();
+   test_type nan_val        = (std::numeric_limits<test_type>::quiet_NaN)();
+   test_type half           = 0.5;
+   test_type one_point_5    = 1.5;
 
    BOOST_CHECK((boost::math::isnormal)(max_val));
    BOOST_CHECK((boost::math::isnormal)(-max_val));
-   BOOST_CHECK((boost::math::isnormal)(min_val));
-   BOOST_CHECK((boost::math::isnormal)(-min_val));
+   BOOST_CHECK((boost::math::isnormal)(denorm_min_val));
+   BOOST_CHECK((boost::math::isnormal)(-denorm_min_val));
    BOOST_CHECK((boost::math::isinf)(inf_val));
    BOOST_CHECK((boost::math::isinf)(-inf_val));
    BOOST_CHECK((boost::math::isnan)(nan_val));
    BOOST_CHECK((boost::math::isnan)(-nan_val));
 
-   if (std::numeric_limits<test_type>::has_denorm)
-      min_val = std::numeric_limits<test_type>::denorm_min();
+   if (!std::numeric_limits<test_type>::has_denorm) {
+      BOOST_CHECK_EQUAL(denorm_min_val, min_val);
+   } else {
+      BOOST_CHECK_LE(denorm_min_val, min_val);
+   }
 
    // Adding epsilon will increment 1.0:
    BOOST_CHECK(test_type(1) + eps != test_type(1));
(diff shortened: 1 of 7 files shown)


ckormanyos:
Great PR it is awesome to see the formal correctness of the types get _even_ closer to mimicing built-ins.

Thanks Ryan (@ryanelandt)

@jzmaddock I can add super-simple test case(s) for these if you think that's a good idea and handle this PR. Thoughts?

jzmaddock:
> @jzmaddock I can add super-simple test case(s) for these if you think that's a good idea and handle this PR. Thoughts?

Yes please!  My apologies for being AWAL recently, hopefully I'll start catching up soon...

A trivial change to test_numeric_limits.cpp would do it.... we also need to fix up at least gmp.hpp and mpfr.hpp which have the same mistakes.

ckormanyos:
> trivial change to test_numeric_limits.cpp would do it.... we also need to fix up at least gmp.hpp and mpfr.hpp which have the same mistakes

Great. I can try this (if I may), ... get back into the game. Some of the recent stuff around here has been a bit too challenging for me. This one seems like a good one I could try.

Cc: @jzmaddock and @ryanelandt

ryanelandt:
@ckormanyos feel free to make the necessary changes.

ckormanyos:
Hi John (@jzmaddock) and Ryan (@ryanelandt) before I get started adding tests and the other two backends, a question, ...

Does the presence of `denorm_min()` mean that `has_denorm` needs to be changed to something _other_ _than_ `std::denorm_absent`?

ckormanyos:
>  presence of (now or soon-to-be) non-zero `denorm_min()` mean that `has_denorm` needs to be changed

Ahhh OK, it is clear from the original post. My bad, ... it is clear from the description use `has_denorm` as _absent_ and then `denorm_min()` is `min()`.

ckormanyos:
Hi Ryan (@ryanelandt)
- The changes also to MPFR and GMP are in and I also added a few basic limits tests.
- CI is cycling.

Cc: @jzmaddock

ckormanyos:
>  changes also to MPFR and GMP

And `mpfi`. I had also forgotten a few needed `(min)()` values. CI is cycling again and looking better...

ckormanyos:
OK we have the change set, it's green on the new tests.

Ryan (@ryanelandt) and John (@jzmaddock) if you see anything i missed or suggest more testing depth, I'd be happy to go another round. Otherwise, form my side, this one is ready to merge.

ckormanyos:
Ryan (@ryanelandt) thanks again and super good catch!

ryanelandt:
Thank you for completing this PR for me. I think it looks good. I have a few questions about the unit tests.

1. Why use `BOOST_TEST(a == b)` instead of `BOOST_CHECK_EQUAL(a, b)`? Is this to avoid needing to have print support for the quantities being tested?

2. Could for example:
```
BOOST_TEST(!(std::numeric_limits<Number>::denorm_min() > (std::numeric_limits<Number>::min)()));
BOOST_TEST(!(std::numeric_limits<Number>::denorm_min() < (std::numeric_limits<Number>::min)()));
```
Be replaced with:
```
BOOST_TEST(std::numeric_limits<Number>::denorm_min() == (std::numeric_limits<Number>::min)());
```
The latter in shorter, it also gives the expected result in the cases that `(std::numeric_limits<Number>::min)()` is `NaN`.

3. Is the check that:
```
BOOST_TEST(FP_ZERO != (boost::math::fpclassify)(std::numeric_limits<Number>::denorm_min()));
```
necessary? Because of the check 2 lines above it, I think it's effect is to check if `BOOST_TEST(FP_ZERO != FP_NORMAL)`.

ryanelandt:
BTW, based on local tests, this PR breaks [this test](https://github.com/boostorg/math/blob/dc10fc04b29a356c3cf12c2052c45b0c3b3da5f2/test/test_autodiff_7.cpp#L60C5-L61C75) in Boost Math.

ckormanyos:
> BTW, based on local tests, this PR breaks

Then we can't merge it. I can look tomorrow. Maybe we need help from Matt (@mborland)?

ryanelandt:
I made a [fix for the Boost Math side](https://github.com/boostorg/math/pull/1015).

ckormanyos:
> made a https://github.com/boostorg/math/pull/1015

Cool. I'd say handle the fix in Math, then we return here. Thanks

test/test_cpp_bin_float.cpp
```diff
@@ -89,24 +93,24 @@ void test_special_cases()
    BOOST_CHECK_EQUAL(max_val * -2, -inf_val);
    BOOST_CHECK_EQUAL(max_val / half, inf_val);
    BOOST_CHECK_EQUAL(max_val / -half, -inf_val);
-   BOOST_CHECK_EQUAL(max_val / min_val, inf_val);
-   BOOST_CHECK_EQUAL(max_val / -min_val, -inf_val);
+   BOOST_CHECK_EQUAL(max_val / denorm_min_val, inf_val);
```
jzmaddock, line 96: I would have both the new *and* old tests here.  Note that the new tests would have passed before (divide by zero).

test/test_cpp_bin_float.cpp
```diff
@@ -89,24 +93,24 @@ void test_special_cases()
...
    BOOST_CHECK_EQUAL(-max_val + max_val, 0);
    // Things which should not over/underflow:
-   BOOST_CHECK_EQUAL((min_val * 2) / 2, min_val);
+   BOOST_CHECK_EQUAL((denorm_min_val * 2) / 2, denorm_min_val);
    BOOST_CHECK_EQUAL((max_val / 2) * 2, max_val);
-   BOOST_CHECK_GE((min_val * 2.0000001) / 1.9999999999999999, min_val);
+   BOOST_CHECK_GE((denorm_min_val * 2.0000001) / 1.9999999999999999, denorm_min_val);
    BOOST_CHECK_LE((max_val / 2.0000001) * 1.9999999999999999, max_val);
-   BOOST_CHECK_EQUAL(min_val * 2 - min_val, min_val);
+   BOOST_CHECK_EQUAL(denorm_min_val * 2 - denorm_min_val, denorm_min_val);
```
jzmaddock, line 113: From lines 99-109 again I would test both min and denorm_min?

test/test_cpp_bin_float.cpp
```diff
@@ -210,26 +214,26 @@ void test_special_cases()
...
-   BOOST_CHECK_EQUAL(max_val + min_val, max_val);
-   BOOST_CHECK_EQUAL(min_val + max_val, max_val);
-   BOOST_CHECK_EQUAL(max_val - min_val, max_val);
-   BOOST_CHECK_EQUAL(min_val - max_val, -max_val);
+   BOOST_CHECK_EQUAL((denorm_min_val / half) * half, denorm_min_val);
+   BOOST_CHECK_EQUAL((denorm_min_val * 2) / 2, denorm_min_val);
+   BOOST_CHECK_EQUAL(max_val + denorm_min_val, max_val);
+   BOOST_CHECK_EQUAL(denorm_min_val + max_val, max_val);
+   BOOST_CHECK_EQUAL(max_val - denorm_min_val, max_val);
+   BOOST_CHECK_EQUAL(denorm_min_val - max_val, -max_val);
```
jzmaddock, line 222: Same again.

test/test_cpp_bin_float.cpp
```diff
@@ -210,26 +214,26 @@ void test_special_cases()
...
+   BOOST_CHECK(boost::math::signbit(-denorm_min_val * denorm_min_val));
+   BOOST_CHECK(boost::math::signbit(denorm_min_val / max_val) == 0);
+   BOOST_CHECK(boost::math::signbit(denorm_min_val / -max_val));
+   BOOST_CHECK(boost::math::signbit(-denorm_min_val / -max_val) == 0);
+   BOOST_CHECK(boost::math::signbit(-denorm_min_val / max_val));
+   BOOST_CHECK(boost::math::signbit(denorm_min_val / 2) == 0);
+   BOOST_CHECK(boost::math::signbit(denorm_min_val / -2));
+   BOOST_CHECK(boost::math::signbit(-denorm_min_val / -2) == 0);
+   BOOST_CHECK(boost::math::signbit(-denorm_min_val / 2));
+   test_type neg_zero = denorm_min_val * -denorm_min_val;
```
jzmaddock, line 236: And again... sorry!!

test/test_numeric_limits.cpp
```diff
@@ -121,6 +121,15 @@ void test_specific(const std::integral_constant<int, boost::multiprecision::numb
          BOOST_TEST(0 != (std::numeric_limits<Number>::min)() / 2);
          BOOST_TEST(0 != std::numeric_limits<Number>::denorm_min());
       }
+      BOOST_IF_CONSTEXPR(std::numeric_limits<Number>::has_denorm == std::denorm_absent)
+      {
+         BOOST_TEST(std::numeric_limits<Number>::denorm_min() > 0);
+         BOOST_TEST(!(std::numeric_limits<Number>::denorm_min() > (std::numeric_limits<Number>::min)()));
+         BOOST_TEST(!(std::numeric_limits<Number>::denorm_min() < (std::numeric_limits<Number>::min)()));
```
jzmaddock, line 128: I would check for equality here?

jzmaddock:
Thanks Chris and Ryan!!

Other than a few minor comments above, this looks good to me.

ckormanyos:
> Other than a few minor comments above, this looks good to me

Thank you John.

This is a silly question, one that I've had for a while. Does anyone know how to simply accept all the suggested comments as a batch change that leads to a single renewed CI run? Or do I/we have to manually change all the codes? Rookie question, I know.

ckormanyos:
OK I can handle all these revies suggestions. Please give me a day or so to do the editing and run CI.

BTW, what is the reason that some of the sets of tests in `test_cpp_bin_float.cpp` seem to be disabled for ASAN and UBSAN?

test/test_cpp_bin_float.cpp
ckormanyos: OK

test/test_cpp_bin_float.cpp
ckormanyos: OK

test/test_cpp_bin_float.cpp
ckormanyos: OK

test/test_cpp_bin_float.cpp
ckormanyos: OK

test/test_numeric_limits.cpp
ckormanyos: I would go for all three of not-`.LT.`, not-`.GT.` _and_ is-`.EQ.`.

ryanelandt:
@ckormanyos there is a lot of code duplication testing things for both `denorm_min` and `min_val`. You might consider putting these tests in a lambda function that looks something like:
```c++
const auto fn_small_val_tests = [&](const T x){
  // Shared tests go here using x instead of `denorm_min` and `min_val`
};

fn_small_val_tests(denorm_min);
fn_small_val_tests(min_val);
```

ckormanyos:
> might consider putting these tests in a lambda function

Good catch Ryan, surely would be better. But, ... I've got refactoring goals on basically all files in Math, Multiprecision and in all of Boost for that matter. 

So my basic philosophy in the test files, which are known by us to be very non-modernized, is, ... leave it until we have a chance to do a dedicated refactor in something like a GSoC.

ckormanyos:
The failure is on a clone attempt on drone. Merging.

ryanelandt:
@ckormanyos Thanks for getting this PR in!

Issue #567, opened 2023-09-21
boost failed to runtest with error C2027, C2676 and error C2059


Zhaojun-Liu:
Recently we updated the boost commit, the below tests failed with `error C2027: use of undefined type 'boost::array<boost::array<T,3>,351>'` and `error C2676: binary '[': 'const boost::array<boost::array<T,3>,351>' does not define this operator or a conversion to a type acceptable to the predefined operator` and `error C2059: syntax error: ')'`
```
...failed compile-c-c++ F:\gitP\boostorg\boost\out\amd64rel\boost\bin.v2\libs\multiprecision\test\test_cos_cpp_dec_float.test\msvc-14.3\release\link-static\runtime-link-static\threading-multi\test_cos.obj...
...failed compile-c-c++ F:\gitP\boostorg\boost\out\amd64rel\boost\bin.v2\libs\multiprecision\test\test_sin_cpp_dec_float.test\msvc-14.3\release\link-static\runtime-link-static\threading-multi\test_sin.obj...
...failed compile-c-c++ F:\gitP\boostorg\boost\out\amd64rel\boost\bin.v2\libs\multiprecision\test\number_concept_check_cpp_dec_float.test\msvc-14.3\release\optimization-space\runtime-link-static\threading-multi\number_concept_check_cpp_dec_float.obj...
...failed compile-c-c++ F:\gitP\boostorg\boost\out\amd64rel\boost\bin.v2\libs\multiprecision\test\number_concept_check_cpp_bin_float.test\msvc-14.3\release\optimization-space\runtime-link-static\threading-multi\number_concept_check_cpp_bin_float.obj...
...failed compile-c-c++ F:\gitP\boostorg\boost\out\amd64rel\boost\bin.v2\libs\multiprecision\test\number_concept_check_backend_concept.test\msvc-14.3\release\optimization-space\runtime-link-static\threading-multi\number_concept_check_backend_concept.obj...
...failed updating 7 targets...
```
**Repro steps:** (Open vs2022 x64 native Tools Command Prompt)
1. git clone https://github.com/boostorg/boost F:\gitP\boostorg\boost
2. git -C "F:\gitP\boostorg\boost" fetch --recurse-submodules=no --force
3. git -C "F:\gitP\boostorg\boost" reset --hard 10994fe
4. git -C "F:\gitP\boostorg\boost" submodule sync
5. git -C "F:\gitP\boostorg\boost" submodule foreach git reset --hard
[...]

PR #568, opened 2023-09-22
Remove uses of Boost.Array

mborland:
Closes: #567


jzmaddock:
Not sure why this worked before... apparently boost::rational has no swap non-member, so we need a using std::swap in the test case.

mborland:
@jzmaddock Any idea why the following now pops up in GNU mode?


```
../../../boost/math/special_functions/beta.hpp:573:35: error: call of overloaded 'log1p(boost::multiprecision::number<boost::multiprecision::backends::mpfr_float_backend<100>, (boost::multiprecision::expression_template_option)0>)' is ambiguous
             power1 = exp(a * log1p((x * b - y * a) / a + p * (x * c / a)));
```

jzmaddock:
That's weird, and new :(

Not just gnu mode as the clang tests are failing too, I'll see if I can reproduce....

jzmaddock:
Fix here: https://github.com/boostorg/math/pull/1043

mborland:
> Fix here: [boostorg/math#1043](https://github.com/boostorg/math/pull/1043)

Thank you. Once that is merged I will re-run the CI here.

Issue #573, opened 2023-12-05
std::has_denorm/std::has_denorm_loss are deprecated in c++23


wanghan02:
`std::has_denorm`/`std::has_denorm_loss` are deprected in c++23. Latest VC++ 19.38 gives warning about this with `/std:c++latest /W4` where we specialize `std::numeric_limits` in `boost::multiprecision`.

https://godbolt.org/z/9Pdfre5Yc

`'std::float_denorm_style': warning STL4042: std::float_denorm_style, std::numeric_limits::has_denorm, and std::numeric_limits::has_denorm_loss are deprecated in C++23. You can define _SILENCE_CXX23_DENORM_DEPRECATION_WARNING or _SILENCE_ALL_CXX23_DEPRECATION_WARNINGS to suppress this warning.`

PR #574, opened 2023-12-06
Add macro to disable MSVC has_denorm deprecation warning

mborland:
Closes: #573

### DIFF
--- a/include/boost/multiprecision/cpp_bin_float.hpp
+++ b/include/boost/multiprecision/cpp_bin_float.hpp
@@ -8,6 +8,12 @@
 #ifndef BOOST_MP_CPP_BIN_FLOAT_HPP
 #define BOOST_MP_CPP_BIN_FLOAT_HPP
 
+#if defined(_MSVC_LANG) && _MSVC_LANG > 202002L
+#  ifndef _SILENCE_CXX23_DENORM_DEPRECATION_WARNING
+#    define _SILENCE_CXX23_DENORM_DEPRECATION_WARNING
+#  endif
+#endif
+
 #include <cmath>
 #include <cstdint>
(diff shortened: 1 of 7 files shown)


jzmaddock:
Are we between a rock and a hard place here?  The problem with _SILENCE_CXX23_DENORM_DEPRECATION_WARNING is that it changes the std lib header and silences the warning everywhere including in user code.  We may also eventually have this issue with other std lib's as well.

I wonder what the correct thing to do is here, members like has_denorm is still required to be present in C++23 but is "a candidate for removal" in a later std.  So we do need to define those members even in C++23 :(  I wonder if there any #pragma warning(disable:XXXX) that would work in this case?

mborland:
> Are we between a rock and a hard place here? The problem with _SILENCE_CXX23_DENORM_DEPRECATION_WARNING is that it changes the std lib header and silences the warning everywhere including in user code. We may also eventually have this issue with other std lib's as well.
> 
> I wonder what the correct thing to do is here, members like has_denorm is still required to be present in C++23 but is "a candidate for removal" in a later std. So we do need to define those members even in C++23 :( I wonder if there any #pragma warning(disable:XXXX) that would work in this case?

We are between a rock and a hard place. I tried disabling the warning, and it still propagates: https://godbolt.org/z/bz1a54jf7

jzmaddock:
Shucks :(

The macro doesn't work either if <limits> has already been included: https://godbolt.org/z/r6Pdn71We

jzmaddock:
The #pragma does seem to work in some situations and not others... will experiment.

jzmaddock:
This seems to get it: https://github.com/boostorg/multiprecision/pull/575

PR #575, opened 2023-12-07
Disable warnings from deprecated numeric_limits members.

jzmaddock:
When building with MSVC in C++23 mode.
Fixes https://github.com/boostorg/multiprecision/issues/573
Refs https://github.com/boostorg/multiprecision/pull/574

### DIFF
--- a/include/boost/multiprecision/cpp_bin_float.hpp
+++ b/include/boost/multiprecision/cpp_bin_float.hpp
@@ -2154,7 +2154,14 @@ class numeric_limits<boost::multiprecision::number<boost::multiprecision::cpp_bi
    static constexpr bool                                                                                                             has_infinity      = true;
    static constexpr bool                                                                                                             has_quiet_NaN     = true;
    static constexpr bool                                                                                                             has_signaling_NaN = false;
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable:4996)
+#endif
    static constexpr float_denorm_style has_denorm                                                                                                      = denorm_absent;
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
    static constexpr bool               has_denorm_loss                                                                                                 = false;
    static number_type infinity()
    {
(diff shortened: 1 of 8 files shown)


include/boost/multiprecision/cpp_bin_float.hpp
```diff
@@ -2154,7 +2154,14 @@ class numeric_limits<boost::multiprecision::number<boost::multiprecision::cpp_bi
...
    static constexpr bool                                                                                                             has_signaling_NaN = false;
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable:4996)
+#endif
    static constexpr float_denorm_style has_denorm                                                                                                      = denorm_absent;
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
    static constexpr bool               has_denorm_loss                                                                                                 = false;
```
mborland, line 2165: Do the instances of `has_denorm_loss` need to go in between the pragmas since it's deprecated as well?

jzmaddock: I don't believe so... it's a simple bool so no warning is generated, it's our use of `float_denorm_style` that was causing the issue I believe.

pps83:
Seems like clang now also complains if c++23 is used.
Something like that fixes the issue:

All the places where this was added:
```c++
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable:4996)
#endif
```
could be changed to:
```c++
#if defined(__clang__) && defined(__has_warning)
# if __has_warning("-Wdeprecated-declarations")
#  pragma clang diagnostic push
#  pragma clang diagnostic ignored "-Wdeprecated-declarations"
# endif
#elif defined(_MSC_VER)
# pragma warning(push)
# pragma warning(disable:4996)
#endif
```

and similary:
```c++
#ifdef _MSC_VER
#pragma warning(pop)
#endif
```
can be changed to:
```c++
#if defined(__clang__) && defined(__has_warning)
# if __has_warning("-Wdeprecated-declarations")
#  pragma clang diagnostic pop
# endif
#elif defined(_MSC_VER)
# pragma warning(pop)
#endif
```

PR #581, opened 2023-12-22
Add codecov run to CI

mborland:


mborland:
@ckormanyos Anything you would add or remove from this?

ckormanyos:
> Anything you would add or remove from this?

Hi Matt (@mborland) I think this is great, both here as well as in  math. And it is really cool to see those coverage numbers above $90\\%%$.

You will chuckle, but I'd kind of like to show the coverage with our codecove-badge. I know some folks do not really like third party badges so much. but I think they add value, unless there is some poicy against them?

Thanks for digging into this Matt´!

ckormanyos:
[codecov](https://codecov.io/gh/boostorg/multiprecision)

ckormanyos:
If you do decide to show the badges, then where? i don't really know. Maybe a third row in the table on the front page Markdown? But we won't have any data from master for a while...

mborland:
> If you do decide to show the badges, then where? I don't really know. Maybe a third row in the table on the front page Markdown? But we won't have any data from master for a while...

I think a third row on the table is the way to go. I'll hold off adding the badges until we have data from master.

PR #585, opened 2024-01-19
Reduce truncation in dec_float multiplication

ckormanyos:
The purpose of this PR is to reduce truncation in `cpp_dec_float`'s multiplication. It is based on some discussion in parts of #368.


ckormanyos:
Hi Matt (@mborland) I can configure the CodeCov Deltas. I'm a bit surprised that my trivial change led to such a decrease. I'll look into this.

Anyway, we dont want fails due to a small one, two percent delta. I can handle this on the weekend in the CodeCov Properties for this workspace.

Issue #586, opened 2024-01-21
Can't convert cpp_dec_float to cpp_int


radj307:
I'm using `cpp_int` and `cpp_dec_float_100` in a project, but I'm encountering an error when trying to convert from float to int on both MSVC 19.34.31947 and gcc 10, using boostmp 1.84 standalone.

```cpp
#include <boost/multiprecision/cpp_int.hpp>
#include <boost/multiprecision/cpp_dec_float.hpp>

namespace mp = boost::multiprecision;

int main()
{
	using mp::cpp_int;
	using cpp_float = mp::cpp_dec_float_100;

	cpp_float r1{ static_cast<cpp_float>(cpp_int{ 5 }) }; //< works
	cpp_int r2{ static_cast<cpp_int>(cpp_float{ 5.5 }) }; //< ERROR: (see below)

	return 0;
}
```

This is the error message in GCC 10 (formatted for readability):

```
operands to ‘?:’ have different types
‘std::enable_if<true, boost::multiprecision::detail::expression<boost::multiprecision::detail::function, boost::multiprecision::detail::number_kind_floating_pointfloor_funct<boost::multiprecision::backends::cpp_dec_float<100> >, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100> >, void, void> >::type’
  {aka ‘boost::multiprecision::detail::expression<boost::multiprecision::detail::function, boost::multiprecision::detail::number_kind_floating_pointfloor_funct<boost::multiprecision::backends::cpp_dec_float<100> >, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100> >, void, void>’}
and
‘std::enable_if<true, boost::multiprecision::detail::expression<boost::multiprecision::detail::function, boost::multiprecision::detail::number_kind_floating_pointceil_funct<boost::multiprecision::backends::cpp_dec_float<100> >, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100> >, void, void> >::type’
  {aka ‘boost::multiprecision::detail::expression<boost::multiprecision::detail::function, boost::multiprecision::detail::number_kind_floating_pointceil_funct<boost::multiprecision::backends::cpp_dec_float<100> >, boost::multiprecision::number<boost::multiprecision::backends::cpp_dec_float<100> >, void, void>’}
(...)\boost\multiprecision\detail\functions\trunc.hpp
```

The error does not occur when using Boost::math alongside it.

[...]

PR #587, opened 2024-01-22
Prelim fix issue586 (if none better found)

ckormanyos:
The purpose of this PR is to provide a minimalistic fix to #586.

In that issue, we see a phenomenon that may require conversion to number-wrap or non-wrapped type detected in overloads of `trunc` and `lltrunc`.

If no better fix is found in this releasce cycle, then this _does the trick_.

Cc: @jzmaddock John take a look if you find better...

Cc: @mborland Standalone with/without the presence of `Math`.

Cc: @radj307

### DIFF
--- a/include/boost/multiprecision/detail/functions/trunc.hpp
+++ b/include/boost/multiprecision/detail/functions/trunc.hpp
@@ -23,23 +23,16 @@ namespace impl {
 template <typename T>
 inline T trunc BOOST_PREVENT_MACRO_SUBSTITUTION (const T arg)
 {
-    using std::floor;
+    if (arg > 0)
+    {
+       using std::floor;
+
+       return floor(arg);
+    }
+
     using std::ceil;
 
-    T t { };
-
-    if (arg > T(0))
-    {
-        t = floor(arg);
-    }
-    else
-    {
-        t = ceil(arg);
-    }
-
-    return t;
-}
-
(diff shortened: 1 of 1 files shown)


ckormanyos:
It's green-ish with CodeCov upload error (re-running) and 1 timeout on drone.

Hi @jzmaddock do you think the potential fix in this PR is sufficient?

Cc: @mborland

jzmaddock:
I'll take a look shortly.

jzmaddock:
That's fine, a simpler fix is:

```
    if (arg > 0) 
       return floor(arg);
    return ceil(arg);
```

which doesn't then require a temporary.

ckormanyos:
> That's fine, a simpler fix is:

Thank you John. It is now cycling in CI.

ckormanyos:
Hi Matt (@mborland) in Multiprecision, uploading Coverage from suite number 1 takes too long and times out all the time, or a lot of the time. We need to eventually look into this.

I can help when we both have a time slot for this...

Cc: @jzmaddock

mborland:
In math every test suite gets its own coverage run because just doing suite 1 and suite 2 also had issues with timeouts. Porting that logic is likely the easiest way to solve this.

ckormanyos:
> every test suite gets its own coverage run

I appreciate that. Many thanks @mborland

PR #596, opened 2024-02-09
Add CodeCov badge for develop

ckormanyos:
Hi @jzmaddock and @mborland I hope you like this as much as I do. I would like quick access and clear visibility for the CodeCov achievements on the front cover ReadMe.

Can we do this? Do we really need table entries for develop/master? Or should i make a separate viewing area for a single CodeCov badge?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,20 @@
 Boost Multiprecision Library
 ============================
 
->ANNOUNCEMENT: This library now requires a compliant C++14 compiler.
+>ANNOUNCEMENT: This library requires a compliant C++14 compiler.
 
 |                  |  Master  |   Develop   |
 |------------------|----------|-------------|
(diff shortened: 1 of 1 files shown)


ckormanyos:
It looks like the following on the page.

grafik

mborland:
I would do one for develop and one for master. Release is in a few weeks so they will both be populated shortly. John is pushing hard on improving math coverage so I think it makes sense to show the delta between the two during the development period.

ckormanyos:
The badge(s) are in.

I also added a (admittedly somewhat subjectively selected) section on _Using_ _Multiprecision_, which I personally would find helpful. It also links to GodBolt.

Like/Disapprove/Don't-Care about the new section? We can retain or modify or delete accordingly. Comments?

Cc: @jzmaddock and @mborland

mborland:
> The badge(s) are in.
> 
> 
> 
> I also added a (admittedly somewhat subjectively selected) section on _Using_ _Multiprecision_, which I personally would find helpful. It also links to GodBolt.
> 
> 
> 
> Like/Disapprove/Don't-Care about the new section? We can retain or modify or delete accordingly. Comments?
> 

I can see it being a useful addition.

PR #601, opened 2024-02-28
Merge for 1.85

mborland:
@jzmaddock, @ckormanyos, and @NAThompson same question here. Any last minute things that need to get added?


jzmaddock:
I suspect the new testing failures are issues in the updated Math coverage tests noy being multiprecision safe.  Running locally now to investigate...

mborland:
> I suspect the new testing failures are issues in the updated Math coverage tests noy being multiprecision safe. Running locally now to investigate...

I will hold off on merging https://github.com/boostorg/math/pull/1104 then.

jzmaddock:
Drone failure is a network timeout, merging...

Issue #602, opened 2024-02-28
`mp_radix_size` pointer type incorrect in some versions of tommath 


mborland:
From the users ML:

I tried to use boost muiltiprecision with Visual Studio 2022, 64 bit and got the following error message:

boost\multiprecision\tommath.hpp(444,36): error C2664: 'mp_err mp_radix_size(const mp_int *,int,size_t *)': cannot convert argument 3 from 'int *' to 'size_t *'

I used tommath.h from

https://github.com/libtom/libtommath/blob/develop/tommath.h

that has (line 583)

      mp_err mp_radix_size(const mp_int *a, int radix, size_t *size) MP_WUR;

but boost\multiprecision\tommath.hpp, lineS 443/444 are

      int s;
      detail::check_tommath_result(mp_radix_size(const_cast< ::mp_int*>(&m_data), base, &s));

I compared with cygwin, there tommath.h is

      mp_err mp_radix_size(const mp_int *a, int radix, int *size) MP_WUR;

So I am not sure: do I need some different tommath for Visual Studio 2022, 64 bit, or should boost\multiprecision\tommath.hpp be changed?

Helmut

mborland:
Change was made: https://github.com/libtom/libtommath/commit/a29aa59baa529d341e2b234ff7331e827aba1ffb

PR #603, opened 2024-02-28
Workaround for type change in tommath

mborland:
Closes: #602 

@jzmaddock This is certainly not the most robust solution on earth, but I see that we check against similar deprecation macros elsewhere. They have no versioning macro so I can't think of a better solution.


jzmaddock:
Thanks Matt, this is about all we can do here I think.

PR #606, opened 2024-03-01
Correct error handling in mpfr specfun overloads.

jzmaddock:


jzmaddock:
Last failure is a network timeout on drone.

@mborland with the changes in Math develop, this now gets Multiprecision develop green and good to go.  Math will need a second merge to master before Multiprecision.

Issue #612, opened 2024-04-15
boost::multiprecision::cpp_int leading zeros parsing bug


ohhmm:
boost::multiprecision::cpp_int("000144") parsed into value 100.
See some details here: https://github.com/ohhmm/openmind/blob/28b9ed5d55bf1766606442cc34689ee0c32671e7/omnn/math/test/Integer_test.cpp#L59

ckormanyos:
>`boost::multiprecision::cpp_int("000144")` parsed into value 100.

Should this parse as an octal base?

mborland:
> > `boost::multiprecision::cpp_int("000144")` parsed into value 100.
> 
> Should this parse as an octal base?

I think anything of the form 0144 should be octal, but 000144 are base-10 leading zeros that need to get stripped out. See the linked PR.

ckormanyos:
> as an octal base?

Because then the expected answer is $8*8+4*8+4=100$.

ohhmm:
Hi. This doesn't look expected to me. Leading zeros hardly should change base of parsing. But since it is C standard compliant, lets close this ticket.

PR #614, opened 2024-04-16
Remove leading zeros from base-10 numbers

mborland:
Currently we assume that anything with a leading 0 is in octal format. Change to where octal has to be 0 followed by a non-zero number. If the format does not meet that requirement for octal than remove any and all leading zeros.

Closes: #612


ckormanyos:
> Change to where octal has to be 0 followed by a non-zero number.

Ah now I get it.

mborland:
@jzmaddock Any issue with this change? It would technically be a breaking change, but I think the new behavior is less surprising.

ohhmm (approved):

jzmaddock:
> Any issue with this change? It would technically be a breaking change, but I think the new behavior is less surprising.

I do worry about this, consider a block of formatted numbers:

```
0001
0002
0003
...
0099
0100
```

We start off as decimal, and then there is a step change to octal when we get to "0100" which seems even worse to me?

mborland:
Godbolt says leading zeros is still octal: https://godbolt.org/z/7Kcahjh67. Since the original ticket is closed I'll just close this too.

ckormanyos:
> start off as decimal, and then there is a step change to octal 

Oh geez I actually had that one last century. It took 3 days to resolve. But now that I reflecrt (which i had entirely forgotten) so is the spec.

Issue #553, opened 2023-05-12
Exception in cpp_int convert_to terminates program


martindolezal:
See the following code: https://godbolt.org/z/cs6cbjnPr

```
#include <boost/multiprecision/cpp_int.hpp>
#include <iostream>

int main() {
    std::string strNumber = "179769313486231590617005494896502488139538923424507473845653439431848569886227202866765261632299351819569917639009010788373365912036255753178371299382143631760131695224907130882552454362167933328609537509415576609030163673758148226168953269623548572115351901405836315903312675793605327103910016259918212890625";
    boost::multiprecision::cpp_int number(strNumber);

    try {
        std::cout << number.convert_to<double>() << "\n";
    } catch (...) {
        std::cout << "some exception caught" << "\n";
    }
    
    std::cout << "finished" << "\n";
}
```

`convert_to<double>()` throws exception here, but it terminates the program, cannot be caught. It seems to be due to `noexcept` qualifier somewhere on the way, probably `eval_convert_to` in cpp_int/misc.hpp

The behavior change between boost 1.78 and boost 1.79.

PR #618, opened 2024-05-03
make sure eval_convert_to() do not terminate with super large number

tchaikov:
this change is a follow-up of d51f2e9dbb. it intends to address the exception thrown in a noexcept functon.

a minimal reproducer looks like

```c++

int main() {
    std::string s = "32767456456456456456545678943512357658768763546575675";
    boost::multiprecision::cpp_int num(s);
    std::cout << num.convert_to<float>() << std::endl;
}
```

since boost 1.79, the code above terminates like
```
Program returned: 139
Program stderr
terminate called after throwing an instance of 'boost::wrapexcept<std::domain_error>'
  what():  Error in function float_next<float>(float): Argument must be finite, but got inf
Program terminated with signal: SIGSEGV
```

because `float_next_imp()` throws 'boost::wrapexcept<std::domain_error>' if the number is NAN of INF. and `eval_convert_to()` is marked as `noexcept(boost::multiprecision::detail::is_arithmetic<R>::value &&
          std::numeric_limits<R>::has_infinity)`,
but only `overflow_error` is ignored in the policy passed to `float_next()`.

so, in this change, `std::domain_error` is ignored as well, so that `num.convert_to<float>()` returns a NaN in this case.

Refs #553

### DIFF
--- a/include/boost/multiprecision/cpp_int/misc.hpp
+++ b/include/boost/multiprecision/cpp_int/misc.hpp
@@ -184,7 +184,9 @@ eval_convert_to(R* result, const cpp_int_backend<MinBits1, MaxBits1, SignType1, 
 
 template <class R, std::size_t MinBits1, std::size_t MaxBits1, cpp_integer_type SignType1, cpp_int_check_type Checked1, class Allocator1>
 inline BOOST_MP_CXX14_CONSTEXPR typename std::enable_if<std::is_floating_point<R>::value && !is_trivial_cpp_int<cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1> >::value, void>::type
-eval_convert_to(R* result, const cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>& backend) noexcept(boost::multiprecision::detail::is_arithmetic<R>::value && std::numeric_limits<R>::has_infinity)
+eval_convert_to(R* result, const cpp_int_backend<MinBits1, MaxBits1, SignType1, Checked1, Allocator1>& backend) noexcept(boost::multiprecision::detail::is_arithmetic<R>::value &&
+                                                                                                                         (std::numeric_limits<R>::has_infinity ||
+                                                                                                                          std::numeric_limits<R>::has_quiet_NaN))
 {
    BOOST_MP_FLOAT128_USING using std::ldexp;
    if (eval_is_zero(backend))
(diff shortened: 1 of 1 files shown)


tchaikov:
a reproducer can be found at https://godbolt.org/z/K5nsj5azT

tchaikov:
hi @jzmaddock , could you help review this change? my colleague @nyh believes that an "inf" should be returned instead, as it was the behaviour in boost 1.78. what do you think?

jzmaddock:
It looks like it's fine: but your parenthesis are in the wrong place around make_policy, should be (I hope):

```
*result = boost::math::float_next(*result, boost::math::policies::make_policy(boost::math::policies::overflow_error<boost::math::policies::ignore_error>(),                                                                                             boost::math::policies::domain_error<boost::math::policies::ignore_error>()));
```

jzmaddock:
Never mind, I see I can edit, have changed thing round...

jzmaddock:
Failures should be corrected by https://github.com/boostorg/math/pull/1130 once it runs...

tchaikov:
> Never mind, I see I can edit, have changed thing round...

thank you! squashed and repushed.

nyh:
> hi @jzmaddock , could you help review this change? my colleague @nyh believes that an "inf" should be returned instead, as it was the behaviour in boost 1.78. what do you think?

I stand by my belief that when converting an integer to floating-point, truncation should result in +Inf (or -Inf) and not NaN. Returning Inf in the case of truncation, and not NaN, is traditional, for example:
```
>>> print(1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.0)
inf
```
but even more importantly, this also this was the behaviour in older Boost before this bug happened, and application code may rely on this truncation behavior for its own user-visible truncation (as for example happened in the ScyllaDB project).

So although after this patch the conversion no longer crashes (good), it's still wrong in my view.

jzmaddock:
My bad, you're quite correct, see https://github.com/boostorg/math/issues/1132.

nyh:
> My bad, you're quite correct, see [boostorg/math#1132](https://github.com/boostorg/math/issues/1132).

Thanks. I'm not familiar with Boost's unit tests, but if you have them, I recommend that you add regression tests for convert_to() whose results are expected to be both +inf **and** -inf, because it's possible the details of what broke (next? prior? I didn't understand the details) would be different in the two cases.

PR #630, opened 2024-07-08
Add ppc64le to CI and fix node20 breakage

mborland:


mborland:
@jzmaddock, @ckormanyos, @NAThompson before I merge this in do you see adding PPC64LE to our CI as valuable? The tradeoff is that it increases the CI run time to about 4 hours. It's not a super high demand architecture but I know in math, charconv, and others there are occasional bug reports.

mborland:
> I guess it's main use is as a "different" architecture. Other than that I'm guessing is basically obsolete now?

Yes it's little endian with two different 128-bit long double ABIs (__ibm128 and __float128) depending on which OS/compiler you are using. I don't think it's obsolete just really expensive. I got a quote for what it would cost to run natively in Drone like ARM and s390x, and it was prohibitive.

jzmaddock:
> Yes it's little endian with two different 128-bit long double ABIs (__ibm128 and __float128) depending on which OS/compiler you are using. I don't think it's obsolete just really expensive. I got a quote for what it would cost to run natively in Drone like ARM and s390x, and it was prohibitive.

Understood.  Many many moon ago, when some hardware vendors provided "compiler farms" you could telnet into, I wondered if it wouldn't be possible to persuade some of them to donate some older machines to the project... but then I realised that someone would have to actually make space for them, put up with the noise and the heat, and feed them actual electrons!

If we (or rather Vinnie) were going to spend actual $$'s, then I suspect providing a CUDA-enabled test machine might be the thing to spend it on.  It's also rather niche, but increasingly important IMO?

mborland:
> > Yes it's little endian with two different 128-bit long double ABIs (__ibm128 and __float128) depending on which OS/compiler you are using. I don't think it's obsolete just really expensive. I got a quote for what it would cost to run natively in Drone like ARM and s390x, and it was prohibitive.
> 
> Understood. Many many moon ago, when some hardware vendors provided "compiler farms" you could telnet into, I wondered if it wouldn't be possible to persuade some of them to donate some older machines to the project... but then I realised that someone would have to actually make space for them, put up with the noise and the heat, and feed them actual electrons!
> 
> If we (or rather Vinnie) were going to spend actual $$'s, then I suspect providing a CUDA-enabled test machine might be the thing to spend it on. It's also rather niche, but increasingly important IMO?

I can ask about some CUDA machines as that would be quite useful. You also can't emulate CUDA so it has to run native.

PR #636, opened 2024-08-02
Make float128 a trivially copyable type

### LINKED ISSUE #635: Is boost::multiprecision::float128 a trivial type?
Hi,

Could you please post the code you used to generate the following?

https://github.com/boostorg/multiprecision/blob/0f6a276756e2fb3b9226000a3b24f788787e2474/example/float128_snips.cpp#L50-L90

For some reason, on my machine, `std::is_trivial<float128>::value` is `false`?

Thank you,
Nuno

nmnobre:
Closes #635.

Opted to go for trivially copyable and not fully trivial as otherwise we'd have to make the `constexpr float128_backend() noexcept` constructor trivial but I expect we'd like to keep zero initialisation for the data member `m_value` we're wrapping even when using default initialisation. In any case, I never needed the type to be trivial, only trivially copyable, so I'm happy if you are?

Code snippet to print type traits and limits: https://godbolt.org/z/ddsj16M7W. I'm not sure we need everything in there, I just copied the original snippet and added a trivially copyable check, but feel free to edit it to your liking and we can then amend the website. :)

### DIFF
--- a/include/boost/multiprecision/float128.hpp
+++ b/include/boost/multiprecision/float128.hpp
@@ -176,12 +176,9 @@ struct float128_backend
 
  public:
    constexpr   float128_backend() noexcept : m_value(0) {}
-   constexpr   float128_backend(const float128_backend& o) noexcept : m_value(o.m_value) {}
-   BOOST_MP_CXX14_CONSTEXPR float128_backend& operator=(const float128_backend& o) noexcept
-   {
-      m_value = o.m_value;
-      return *this;
-   }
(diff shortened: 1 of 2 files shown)


mborland:
Thanks for this. I also added a trivial test: https://github.com/boostorg/multiprecision/commit/c2b11c1e9170686f997b158d368b3c83e3f572a4, and update the code snippet from your original question: https://github.com/boostorg/multiprecision/commit/b16cff21e3bcd5d7e30538578913fb2373c1720a

PR #641, opened 2024-11-04
Disable check for `__float128` on PGI compiler

### LINKED ISSUE #640: nvc++ errors with message "float128_type is undefined"
nVidia's HPC toolkit (https://developer.nvidia.com/hpc-sdk-downloads) supplies compiler nvc++, which defines macro __PGI. Using it to compiling a C++ file that #includes <boost/geometry/index/rtree.hpp> issues the following error message (this was with Boost 1.83.0, but the error began occurring with 1.79.0):

.../boost_1_83_0/boost/multiprecision/detail/number_base.hpp", line 205: error: identifier "float128_type" is undefined
  struct bits_of<float128_type>

Tested operating system: Ubuntu 20.04 LTS
Compiler: nvc++ 23.7-0 64-bit target on x86-64 Linux -tp haswell

mborland:
Fixes #640 

Per config we should not have 128-bit float on this platform: https://github.com/boostorg/config/blob/19701e05d4832d1e97e06e791f2e4e1c7cba9b13/include/boost/config/compiler/pgi.hpp#L19.

Not sure if anything has happened in the 7 years since that commit was last updated...

### DIFF
--- a/include/boost/multiprecision/detail/number_base.hpp
+++ b/include/boost/multiprecision/detail/number_base.hpp
@@ -206,7 +206,7 @@ struct bits_of
                                               : sizeof(T) * CHAR_BIT - (boost::multiprecision::detail::is_signed<T>::value ? 1 : 0);
 };
 
-#if defined(_GLIBCXX_USE_FLOAT128) && defined(BOOST_GCC) && !defined(__STRICT_ANSI__)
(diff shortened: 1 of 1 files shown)


mborland:
The 3 failed run in circle all provide 0 information as to the failure, it's all build killed. Since GitHub actions and drone both show no problems it's likely safe to merge this. The issue reporter also says it fixed their problem.

PR #648, opened 2025-01-06
Integrate cpp_double_fp_backend

ckormanyos:


ckormanyos:
Retry due to confusion:

This PR attempts to add `cpp_double_fp_backend<T>`.

There are still a few TODOs (a very few) in the original GSoC repo.

I've started this PR in the _real_ Boost nonetheless in the hopes that we might make 1.88 (or later 1.89). There's not much happenning in Multiprecision so it does not really matter if this PR is open for 5 weeks or 5 months.

I also want to gain experience with the full CI

Cc: @jzmaddock and @sinandredemption and @cosurgi and @mborland

ckormanyos:
OK. This thing is going well.

The only failures are on x390 and Apple 64 bit on Arithmetic-Tests due to some confusion with `<limits>`. So basically the hard parts are done on `cpp_double_fp_backend<T>`.

I'll figure out how to deal with those non-x86_64 architectures (which I'm happy to have in CI) and finish off this thing.

This is getting close rather sooner than later.

Cc: @sinandredemption and @jzmaddock and @mborland and @cosurgi

cosurgi:
This is seriously awesome. Thank you so much for your great work. I will try to find some time to look at this more closely.

ckormanyos:
OK @sinandredemption and @cosurgi and @jzmaddock the `cpp_double_fp_backend` has gone green. There are some rudimentary docs as well.

I still need to actually _build_ and puch the docs. But I might do this after the upcoming review in another candidate Boost library.

Um from my subjective standpoint, this backend is ready to go out anr run wild. Let's see how Janek's first hard-real-world test comes out, good, neutral or disaster and then see where we want to go with this backend.

I do not have any real huge open issues. There are a few subtle checks I/we could add for overflow catch in add/sub. But at the moment I am hesitant to slow down the backend. It is at the moment 3x faster than the corresponding dec-float backend in my Mandlebrot bench.

And I'm excited to see how this thing faces off with other backends in Janek's Yade simulation.

Let's see how it goes...

Cc: @mborland and @NAThompson

ckormanyos:
See also https://github.com/BoostGSoC21/multiprecision/issues/181

ckormanyos:
OK this thing is green again with the improvements on performance, especially the `str()` method.

I will add the updated performance data to the documentation and add some motivational words. These words will also be cautionary and clearly state that the most powerful benefit of this backend at the moment is for pure numerical calculations and that the internal string-streaming operations are not (yet) optimal.

So I need to:

- Add perf data to the docs.
- Build the docs.

Then this thing is good to go.

Cc: @sinandredemption and @cosurgi and @jzmaddock

ckormanyos:
The coverage on `cpp_double_fp.hpp` could be improved. But this should not impede this pR in my opinion at the moment.

See also https://github.com/BoostGSoC21/multiprecision/issues/182

ckormanyos:
> The coverage on `cpp_double_fp.hpp` could be improved.

So the coverage is, let's say, adequate and is reaching the ${\approx}94\\%$ for the overall project. A few things are found such as the internal function `pown` is not called (at least in coverage.

Also the `float128` stuff is not included in the coverage report, although this is being called. I guess I'd need another job in the coverage YAML to hit these lines. Simple enough but let's see if I get to it.

ckormanyos:
>> The coverage on cpp_double_fp.hpp could be improved.

> So the coverage is, let's say, adequate

I'm cool with it at this juncture. The report has 52 lines missing in coverage from `cpp_double_fp.hpp`. I'm satisfied with that for the first cut.

Next stop, add perf-info, build docs and move forward.

ckormanyos:
> I'm cool with it at this juncture. The report has 52 lines missing in coverage from cpp_double_fp.hpp. I'm satisfied with that for the first cut.

It ought to be simple enough to pep the entire project up to 95-plus coverage when dealing with a few more edge cases in the future. but I'll leave that quality goal for another tmie, ...

ckormanyos:
Hi @jzmaddock and @mborland.

We worked a bit on this one. I'm happy with the functionality, performance and quality. I'm also moderately pleased with the testing, the coverage and the overall correctness and behavior of this backend.

On last thing --- I need to build and check in the docs. But the double-double (or as i so lovingly call it, the double-trouble backend) is good to go.

I'm ready with this thing. Are there any significant comments, reservations or change requests before I doc this thing and push to develop?

This might take a week or 2 in order to finish, during which time this comment phase is fully open.

Cc: @sinandredemption and @cosurgi and @NAThompson

ckormanyos:
This is part one of https://github.com/boostorg/multiprecision/issues/184

And my favorite benchmark (and I'm using this thing already for high-speed fractal iterations). Pure algebra power:

### Pure Algebra Timing Report

- Full classic Mandelbrot image $2048{\times}2048$ pixels at 32 decimal digit precision.
- `cpp_double_double` versus `cpp_dec_float<32>`
- $2000$ iterations, half-width $1.35$, centered around the point $(0.0, -0.75i)$.
- Using 15 cores multithreading on Intel Core-i9.

| number type                                    | time (s)  | factor |
| -------------------------------------- | --------- | -------|
| `cpp_double_double`                       | 17          |[1]
| `cpp_dec_float<32>`                       | 52          |3|

And the picture is, ... beautiful!

mandelbrot_MANDELBROT_01_FULL

cosurgi:
How about Mandelbrot in `float128` :)

ckormanyos:
> How about Mandelbrot in float128 :)

Good question Janek (@cosurgi)! I'll let you know. That thing ran in Win*. But since elementary add/sub/mul/div were always faster on double-trouble i don't expect any problems. But the proof is in the pudding. I'll run on *nix and tll you in a day or so.

ckormanyos:
>> How about Mandelbrot in float128 :)

> Hi Janek, the suspense was killing me, ...

double-trouble 12s
float128 on WSL2 26s

image

image

cosurgi:
Awesome!

ckormanyos:
We had a stumble on yade. But this thing actually performs on raw algebra. We just need to work out the kinks. I did not even optimize it fully yet regarding single-limb, int, etc. operations. So it can get better.

When we started this thing, I was skeptical. Now I'm convinced. If you have a limited exponent range and you just need a few more digits, then double-fp-backend might be a good choice.

I'm excited about quad.

cosurgi:
> I'm excited about quad.

me too! :)

ckormanyos:
But we still need to face yade and see if there are any issues there needing to be addressed. We can do this as soon as you work out reduction of string-streaming.

ckormanyos:
OK so `cpp_double_fp.hpp` is up to almost $99\\%` lin coverage. All tests are passing. Perf on algebra is good. All of Ci is green.

There are 8 missing cover lines in the header. I might go for 3 lines or so. The rest are actually open issues in my mind.

So final steps are:

- Finish coverage.
- Build anc check in docs.

Cc: @sinandredemption and @cosurgi and @jzmaddock

cosurgi:
I have a preliminary version of yade where I eliminate string conversions, the CI is cycling here: https://gitlab.com/yade-dev/trunk/-/merge_requests/1070

Tomorrow I have to focus on something else, but in a couple of days I should be able to run a benchmark of `cpp_double_*` types and see what really is the speedup with this new `cpp_double_fp_backend`.

ckormanyos:
> run a benchmark of `cpp_double_*` types and see what really is the speedup with this new `cpp_double_fp_backend`.

At the moment, I am relatively confident in pure algebraic speed. There are a few open issues on some elementary functions. And read/write string is known inefficient.

So all-in-all the `cpp_double_fp_backend` is ready for something like a _first_ _release_. But it might take a while to ensure that every single part of the backend is efficient. And I expect this will happen in a sequence of fixes and iterations over the next few months.

Let's see where we sand now. Thankis Janek (@cosurgi).

ckormanyos:
OK I'm done with the technical development of `cpp_double_fp_backend`. And I am satisfied with this first draft.

Summary:

- Running in tests and CI and locally for `cpp_double_fp_backend`
- Performance very good for all of algebra and most special functions.
- Missing only 4 lines of coverage so above $99\\%$ cover.

Just need to finish the docs.

Cc: @sinandredemption and @cosurgi and @jzmaddock and @mborland

cosurgi:
Hi Chris (@ckormanyos) this is very hot off the press. I have just managed to reduce the number of `rd_string` calls from 233480 to 48 and `str` calls from 194842 to 437 in the `./examples/yade -n --quickperformance -j 1` benchmark!

I cannot run the timing benchmarks yet, because I need to clean up the code and remove all these `std::cerr << __PRETTY_FUNCTION__ << "\n";`  everywhere ;) They slow down the calculations!

Now if we manage to add `pow` for integer powers we will be good to go!

ckormanyos:
> Hi Chris (@ckormanyos) this is very hot off the press. I have just managed to reduce the number of `rd_string` calls from 233480 to 48 and `str` calls from 194842 to 437 in the `./examples/yade -n --quickperformance -j 1` benchmark!

This is great. Nice work.

> Now if we manage to add `pow` for integer powers we will be good to go!

I have not done that, but OI verified that we go to a somewhat efficient calculation in our existing collection of default functions

Now, we might still face some quirky issue, and the _proof_ will be in your final bench run. If we are speedy, then good. If, for some other reason, we still face slow downs then we have two wins:

- YADE is going to get better
- And if we still need to squeeze more `cpp_double_fp` functions, then we can find them and push their limits down.

So no matter what actual lnumbers you get, your work has put us in a stronger position. Let's keep going!

Cc: @sinandredemption and @jzmaddock

cosurgi:
Wow, with last single small change I reduced the `rd_string` calls to 2 !

cosurgi:
> I have not done that, but I verified that we go to a somewhat efficient calculation in our existing collection of default functions. If this ends up being one of the"functions-to-speed-up", then I can do that fast.

Hi Chris (@ckormanyos), what happens if you replace `z = z*z + c;` with `z = pow(z,2) + c;` in your Mandelbrot benchmark?

ckormanyos:
> Hi Chris (@ckormanyos), what happens if you replace `z = z*z + c;` with `z = pow(z,2) + c;` in your Mandelbrot benchmark?

It ruins the performance completely and entirely. What a great question Janek. It took so long that I am still waiting for the timing result. i had real/imag components separated. See the pic below.

In summary, the pow function killed performance on that particular benchmark.

I went from 17 seconds to 170 seconds, a factor of 10

image

ckormanyos:
Nightmare timing up by a factor of 10.

oops, it is time to hand-optimize `pow(x, n)` where $n$ is an integer.

image

cosurgi:
Whew. That's good news for me actually, because yade uses `pow(arg, 2)` 436138 times in the calculations. And I have removed calls to `rd_string` entirely. And I got this benchmark result:

| type                    | calculation speed     | factor   |
| ----------------------- | --------------------- | -------- |
| float128                | 159.4864 iter/sec     | 1        |
| cpp_double_double       | 31.3289 iter/sec      | 5.09     |

Meaning that in yade `float128` is still 5 times faster than `cpp_double_double`.

Before removing `rd_string` it was like this:

| type                    | calculation speed     | factor   | commit   |
| ----------------------- | --------------------- | -------- | -------- |
| float128                | 145.3411 iter/sec     | 1        |          |
| cpp_double_double       | 30.4207 iter/sec      | 4.77     | 9f346580 |

So you may notice that `float128` performance increased by a factor of 1.1 thanks to removing of `rd_string`. And `cpp_double_double` is also faster, but only a tiny bit: by a factor of 1.02.

cosurgi:
I am not sure if on that screenshot the lines 140 and 141 is correct? You have `zr2 = pow(zr, zr)`, shouldn't it be:`zr2 = pow(zr, 2)` ?

ckormanyos:
> I am not sure if on that screenshot the lines 140 and 141 are correct? You have `zr2 = pow(zr, zr)`,
shouldn't it be: `zr2 = pow(zr, 2)`? 

You are right Janek. That was a silly, late evening, hurried blunder.

When I used the proper `pow(zr, 2)`, the timing was worse $24s$ compared to $17s$, but not so bad as the previous report.

image

jzmaddock:
Just curious, did we not optimize the default pow function for integer exponents?

Also just FYI the boost::math::pow<N>(x) function is designed to optimise exactly this case: a power with a constant integer exponent.  So far as I know there is no way within the language to detect that pow(T, int) is being called with an integer literal?

ckormanyos:
> Just curious, did we not optimize the default pow function for integer exponents?

Yes John, you are right. The generic collection of functions in Multiprecision DOES include specializations of `eval_pow` for pure integral powers.

I am experimenting with a local version, but I am not able to get significantly faster than the default version in Multiprecision, maybe only $10-20\\%$ faster.

At the moment, I am not able to see any more clear bottlenecks in the overall performance of `cpp_double_fp_backend`.

cosurgi:
Chris (@ckormanyos) can you share your Mandelbrot benchmark code? I want to make sure that I can reproduce your results. Because if I don't then we know it's not a problem with `cpp_double_fp` but with my local configuration.

cosurgi:
Chris (@ckormanyos) in [this post](https://github.com/boostorg/multiprecision/pull/648#issuecomment-2585919468) with the Mandelbrot benchmark which version of g++ and optimization flags (`-O3`, `-Ofast` ?) did you use to compare `cpp_double_double` with `float128` ?

ckormanyos:
> the Mandelbrot benchmark

See also: https://github.com/BoostGSoC21/multiprecision/issues/190

Hi Janek (@cosurgi), I have made a dedicated issue for this discussion. In that issue, I will provide the benchmark code and, yes, it does offer the ability to compare bin-float, dec-float, float128 and double-double.

Give me a day or so to prepare a branch of the Mandelbrot for your dedicated use.

Cc: @jzmaddock and @sinandredemption

ckormanyos:
> does this PR improve power performance at all

Hi John (@jzmaddock). In a word, yes. Treating small powers in that super-fast way is something we should probably do.

Another thing I have been playing around with is a more subtle issue. In my recent pushes here, I have introduced a concept called `mul_unchecked`. And this just cacled green.

In `mul_unchecked` I skip the proölogue to multiplication which checks for NaN, infinity, zero and the like. Sort of making a pure multiplication that is separate from the `eval_mul` operation.

As it turns out, the floating-point-class checks actually do slow down these tiny, tiny backends significantly. We also found this to be relevant for the work in decimal. So I will also be sepüarating the work from the safety of mul/div a bit down the road as well.

As for your changes there, I think they definitely help all of multiprecision, but I still _might_ end up specializing $x^n$ for the double-float backend if that squee<zes out $5\%%$ or more, as it seems to in my recent studies.

Cc: @cosurgi

cosurgi:
I have finally some good news! Here's the latest YADE benchmark, `yade -n --quickperformance -j 4`. I am working with branch cpp_double_fp_backend commit c6ce52dacc3ce.

The trick was to use a different compiler!

clang++ 19.1.4, `-O3`

| type                                     | calculation speed     | factor   |
| ---------------------------------------- | --------------------- | -------- |
| `cpp_double_double` clang++ 19.1.4         | 205.0179 iter/sec     | 1        |
| `cpp_bin_float<32>` clang++ 19.1.4         |  95.3581 iter/sec     | 2.14     |
| `cpp_dec_float<31>` clang++ 19.1.4         |  49.1410 iter/sec     | 4.17     |
| `mpfr_float_backend<31>` clang++ 19.1.4    |  31.7974 iter/sec     | 6.44     |

g++ 14.2, `-O3` (compared to clang `cpp_double_double`)

| type                                     | calculation speed     | factor   |
| ---------------------------------------- | --------------------- | -------- |
| `float128`   g++ 14.2                      | 165.5817 iter/sec     | 1.23     |
| `cpp_bin_float<32>` g++ 14.2               |  98.4807 iter/sec     | 2.08     |
| `cpp_dec_float<31>` g++ 14.2               |  51.2811 iter/sec     | 3.99     |
| `cpp_double_double` g++ 14.2               |  34.2752 iter/sec     | 5.98     |
| `mpfr_float_backend<31>` g++ 14.2          |  30.6851 iter/sec     | 6.68     |

So all other types perform at nearly the same speeds for both compilers. The only exception is `cpp_double_double` which has a huge difference in performance: **`cpp_double_double` is 6 times slower with g++** ! Is that a problem with g++ ? What else could it be?

So Chris, if you used a different compiler for your Mandelbrot `float128` benchmark, it may explain the discrepancies. But AFAIK only g++ supports `float128` so maybe there is still something to explain. But at least I got some good results from clang!

I wanted to post here, because the news are good. But yeah, we can continue in https://github.com/BoostGSoC21/multiprecision/issues/190

ckormanyos:
Note to self: TODO Hit the edge-cases of the new `eval_pow` method.

ckormanyos:
Performance of algebraic functions re-affirmed in https://github.com/BoostGSoC21/multiprecision/issues/190

cosurgi:
OK, so the bad performance mystery was [solved](https://github.com/BoostGSoC21/multiprecision/issues/190#issuecomment-2599763610) and I did  benchmarks of [YADE software](https://yade-dem.org/doc/) `yade -n --quickperformance -j 4` on a quite recent CPU Intel i7-14700KF and the results are good. Some are interesting. We can definitely mark the performance problem of the `cpp_double_fp_backend` as solved. Now only the compiler writers will have something to talk about :)

Here are the results:

### `cpp_double_double`

| type                                     | calculation speed     | factor   |
| ---------------------------------------- | --------------------- | -------- |
| `cpp_double_double`       g++ 12.2       |   449.15 iter/sec     |   1      |
| `float128`                g++ 12.2       |   263.15 iter/sec     |   1.70   |
| `cpp_bin_float<32>`       g++ 12.2       |   211.81 iter/sec     |   2.12   |
| `cpp_dec_float<31>`       g++ 12.2       |    78.15 iter/sec     |   5.74   |
| `mpfr_float_backend<31>`  g++ 12.2       |    51.01 iter/sec     |   8.80   |

Here we can see that `cpp_double_double` beats everyone else by over a factor of two.

### `cpp_double_long_double`

| type                                     | calculation speed     | factor   |
| -----------------------                  | --------------------- | -------- |
| `cpp_bin_float<39>`       g++ 12.2       |   122.55 iter/sec     |   1      |
| `cpp_double_long_double`  clang++ 19.1.4 |   108.79 iter/sec     |   1.12   |
| `cpp_bin_float<39>`       clang++ 19.1.4 |   102.19 iter/sec     |   1.20   |
| `cpp_dec_float<39>`       g++ 12.2       |    71.42 iter/sec     |   1.71   |
| `mpfr_float_backend<39>`  g++ 12.2       |    45.75 iter/sec     |   2.67   |
| `cpp_double_long_double`  g++ 12.2       |    14.97 iter/sec     |   8.18   |

Here we can see that `cpp_double_long_double` performs very good. But the compiler writers will have a mystery to solve: `cpp_bin_float<39>`       g++ 12.2 is faster than `cpp_double_long_double`  clang++ 19.1.4 by just a little, which in turn is faster than `cpp_double_long_double`  g++ 12.2 by a factor of 8.

### `cpp_double_float128`

| type                                     | calculation speed     | factor   |
| -----------------------                  | --------------------- | -------- |
| `cpp_bin_float<67>`   g++ 12.2                     |   118.43 iter/sec     |   1      |
| `mpfr_float_backend<67>`      g++ 12.2             |    43.34 iter/sec     |   2.73   |
| `cpp_dec_float<67>`          g++ 12.2              |    40.09 iter/sec     |   2.95   |
| `cpp_double_float128`      g++ 12.2                |    14.99 iter/sec     |   7.90   |

Here we can see that `cpp_double_float128` has a lot of potential to beat `cpp_bin_float<67>` once the g++ developers sort out the problems with `cpp_double_long_double`  g++ 12.2. The increase in performance should be about by a factor of 8 :)

So all is good. I think we can merge this branch once documentation and other small TODOs are complete.

ckormanyos:
> We can definitely mark the performance problem of the `cpp_double_fp_backend` as solved.

Thank you Janek (@cosurgi) that was a big effort, and it really provided a lot of information and clarity.

Some of the results on `cpp_double_long_double`, where `long double` is 80-bit, 10-byte in width are interesting. That hardware version of the 10-byte floating-point representation is running on the legendary (modernized) versions of the i387 FPU, the hardware that really put 10-byte floating-point on the map.

The newer i7 processors have extremely powerful 64-bit floating-point hardware operations, and it seems like these are being very well supported nowdays in hardware and software.

Down the road I will be doing some non-x86_64 measurements on M1 and/or M2 and a few embedded bare-metal controllers like  an ARM(R) Cortex(R) M7, having double-precision floating-point FPU support.

All-in-all I'm somewhat surprised at how fast `cpp_double_double` ended up in certain harware/software configurations. As mentioned in precious posts, this backend (and of course that type specifically) have lots of room for optimization improvement.

Cc: @sinandredemption and @jzmaddock

jzmaddock:
There *might* be one more thing to check: that each of the backend/compiler configurations are doing (roughly) the same amount of work.  Something that can happen when there is a tolerance set for termination is you can hit "unfortunate" parameters which cause the code to thrash through many needless iterations which don't actually get you any closer to the end result.  I have no idea if this is the case here, but because they don't behave quite like exactly rounded IEEE types, things like `double double` can easily break assumptions present in the code.

ckormanyos:
> There might be one more thing to check: that each of the backend/compiler configurations are doing (roughly) the same amount of work. Something that can happen when there is a tolerance set for termination is you can hit "unfortunate" parameters which cause the code to thrash through many needless iterations which don't actually get you any closer to the end result.

Indeed. There are several potential dangers.

Let's say we use `cpp_double_double` and a particular tolreance $dx$ is set to

$$
{\vert|}dx{\vert|} < 1^{-300}
$$

At the same time, we know that `min_exponent10` for the type is something like $-291$. So the tolerance is never reached or reached after useless iterations.

Even worse, this backend is new, so there might be undiscovered problems in the areas of subnormal/zero. So you might iterate until the maximum iteration setting.

We actually had several cases like this when John helped me _see_ through the last tricky spots in the _specfun_  tests. Who knows if we really got all the edge cases?

Cc: @jzmaddock and @cosurgi

ckormanyos:
So I have updated this PR to the post-1.88 develop branch of Multiprecision. And it's going green once again (with the updated CI). There were a few trip-ups along the way, but nothing out of the ordinary.

It's time to continue working the known final points in
[BoostGSoC21/multiprecision/issues/160](https://github.com/BoostGSoC21/multiprecision/issues/160).

I'm not sure if all this will be ready for 1.89, but there is a chance.

Cc: @cosurgi and @jzmaddock and @sinandredemption

ckormanyos:
Hi @LegalizeAdulthood thank you for your review points. I will get to these. Your points seems sensible.

It's been a while in the making, but I use this backend already locally and it seriously accelerates perturbative Mandelbrot calculations like by a factor of 3. no other backend kicks it like this one at double-double for about 32 digits.

I do not know if7when I'll get this Biist-ready, but I#m still on it.

Cc: @jzmaddock

ckormanyos:
Hi @LegalizeAdulthood I'll leave all the conversations open for now. I need to get back to these a bit later. Thanks again for contributing.

ckormanyos:
> If I can be of assistance in helping this pull request be accepted, let me know. 

Hi Richard (@LegalizeAdulthood) if you get a chance, you could consider using the `cpp_double_fp_backend` backend. Boost.Multiprecision is header-only. So if you checkout the `cpp_double_fp_backend_integration` branch, you can immediately use classes such as `boost::multiprecision::cpp_double_double`.

Other than that, I think there are still some edge cases in rounding and round-tripping. Those are the only open points left. I deactivated many of the tests in rounding and round-tripping and I think these _should_ pass. I'll need to talk with John about these sometime down the road.

ckormanyos:
Redundant with #515

PR #649, opened 2025-01-17
Try and optimize small integer powers.

jzmaddock:


ckormanyos:
Thank you John!

cosurgi:
Hi John (@jzmaddock), if it's possible please add a `default:` case which does nothing. I am compiling YADE with flags (among others) `-Werror -Wswitch-default`. This will produce a compilation error when `default:` is missing

PR #654, opened 2025-02-13
Add constexpr support for dynamic allocated cpp_int

marcoffee:
Added constexpr support for dynamic allocated `cpp_int`s as long as compiler implements features `__cpp_constexpr_dynamic_alloc` and `__cpp_lib_constexpr_dynamic_alloc` from [P0784R7](https://wg21.link/P0784R7).

If those features are disabled, it behaves as the current implementation. Note that those features require C++20, so they will only work when passing `-std=c++2a` (or beyond) flag during compilation through a supported compiler (tested here on `g++-11`+ and `clang++-16`+ using libstdc++ with `g++-10`+ toolset on ubuntu noble).

It might work on `g++-10`, but for some reason the b2 check for `if constexpr` keep returning "no" for it.

Also fixed constexpr construction from strings and added some tests for construction/operations on dynamic allocated `cpp_int`s.

Had to make small changes to existing code to tackle some warnings given when compiled without the feature's support (such as uninitialized variables not being supported in constexpr methods before C++20).


ckormanyos:
Hi @marcoffee I'm sorry this took so long. Your PR was waiting in the queue for a while. I've just approved the CI/CD run on this PR. In Multiprecision, first-time contributors need approval to even _run_ CI/CD, and this can be a bit of a point where we lose some time. Sorry for that.

There is renewed interest in `constexpr`-ification of `cpp_int` with allocator storage.

Of course, @jzmaddock will need to look at this, and I will also. But let's see how CI/CD runs with the full battery of compilers running it.

Cc: @jzmaddock and @mborland and @eisenwave

ckormanyos:
Hi @marcoffee and @jzmaddock CI has run green. One thing, however, that I did not test is a simple code verifying the `constexpre`-ness of an allocator-based `cpp_int`.

@marcoffee could you work up an  example (post-C++20) that exercises this? Then we can put the test into CI/CD.

Cc: @mborland

ckormanyos:
The added missing lines in code coverage seem to be lines that aare never expected to run, using `BOOST_MP_IS_CONST_EVALUATED`. I am not too (perhaps naively) concerned about these?

ckormanyos:
I can't merge this thing until we get a consensus from @jzmaddock and @mborland

mborland:
> The added missing lines in code coverage seem to be lines that aare never expected to run, using `BOOST_MP_IS_CONST_EVALUATED`. I am not too (perhaps naively) concerned about these?

It's not that they are never used, it's that they never used at run-time so codecov can't pick them up. I normally just `LCOV_EXCL_LINE` everything inside these kinds of `if consteval` blocks.

marcoffee:
@ckormanyos thanks for your reply!

> @marcoffee could you work up an example (post-C++20) that exercises this? Then we can put the test into CI/CD.

Please correct me if I'm wrong, but some of the tests (e.g., [here](https://github.com/marcoffee/multiprecision/blob/feature/more-constexpr/test/constexpr_test_dynamic_cpp_int.hpp#L159)) can only run during constexpr evaluation since they are called in a `static_assert`. Since we use an infinite `cpp_int` type, they are also dynamically allocated.

If there is need for more tests I can add them. Also, if there is need for adding some `LCOV_EXCL_LINE` or `LCOV_EXCL_START`/`LCOV_EXCL_STOP` blocks I could look into adding them.

mborland:
> @ckormanyos thanks for your reply!
> 
> > @marcoffee could you work up an example (post-C++20) that exercises this? Then we can put the test into CI/CD.
> 
> Please correct me if I'm wrong, but some of the tests (e.g., [here](https://github.com/marcoffee/multiprecision/blob/feature/more-constexpr/test/constexpr_test_dynamic_cpp_int.hpp#L159)) can only run during constexpr evaluation since they are called in a `static_assert`. Since we use an infinite `cpp_int` type, they are also dynamically allocated.
> 

I looked through the tests this morning and yes, it looks like you're already exercising these dynamic allocation paths. 

> If there is need for more tests I can add them. Also, if there is need for adding some `LCOV_EXCL_LINE` or `LCOV_EXCL_START`/`LCOV_EXCL_STOP` blocks I could look into adding them.

If you would add `LCOV_EXCL_LINE` or `LCOV_EXCL_START`/`LCOV_EXCL_STOP` to blocks that look like this:

```cpp
if (BOOST_MP_IS_CONST_EVALUATED(n))
{
    // LCOV_EXCL_START

   ...

    // LCOV_EXCL_STOP
}
```

Or `LCOV_EXCL_LINE` if it's a one liner.

Codecov will erroneously mark these paths as never taken, but it doesn't support coverage of constant evaluated paths.

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -63,6 +64,91 @@ namespace detail {
...
+   template <class Allocator>
+   static BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR
+   typename std::allocator_traits<Allocator>::pointer constexpr_allocate_default_constructible(Allocator& alloc, std::size_t const n)
+   {
+      typename std::allocator_traits<Allocator>::pointer ptr = alloc.allocate(n);
+
+#     if !defined(BOOST_MP_NO_CONSTEXPR_DETECTION) && defined(BOOST_MP_HAS_CONSTEXPR_DYNAMIC_ALLOC)
+      if (BOOST_MP_IS_CONST_EVALUATED(n))
+         for (std::size_t i = 0; i < n; ++i)
+            std::construct_at(ptr + i);
```
mborland, line 119: Could you please add braces to these? You can quickly get into trouble with these nested

config/has_constexpr_dynamic_memory.cpp
```diff
@@ -0,0 +1,25 @@
+//  Copyright John Maddock 2025.
```
mborland: You can copyright all of these into your own name instead of John's

include/boost/multiprecision/cpp_int/limits.hpp
```diff
@@ -61,103 +64,124 @@ get_min(const std::integral_constant<bool, true>&, const std::integral_constant<
...
 get_max(const std::integral_constant<bool, false>&, const std::integral_constant<bool, true>&, const std::integral_constant<bool, has_allocator>&)
 {
-   // Unbounded and signed.
+   // Unbounded and signed, can be constexpr if allocations are constexpr.
    // There is no maximum value, just return 0:
-   static const boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates> val(0u);
+#ifdef BOOST_MP_NO_CXX20_DYNAMIC_ALLOC_CONSTEXPR_DETECTION
+   static
+#endif
+   const boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates> val(0u);
```
mborland: What prevents these from being `static` in the constexpr case?

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -427,67 +521,73 @@ struct cpp_int_base<MinBits, MaxBits, signed_magnitude, Checked, Allocator, fals
...
    {
       if (!m_internal && !m_alias)
-         allocator().deallocate(limbs(), capacity());
+         detail::constexpr_deallocate_trivially_destructible(allocator(), limbs(), capacity());
    }
-   void assign(const cpp_int_base& o)
+   BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR void assign(const cpp_int_base& o)
    {
-      if (this != &o)
+      if (static_cast<const void*>(this) != static_cast<const void*>(&o))
```
mborland: Why do we now need to cast these to `const void*`?

test/constexpr_test_dynamic_cpp_int.hpp
```diff
@@ -0,0 +1,494 @@
...
+#include "boost/multiprecision/cpp_int.hpp"
+#include "test.hpp"
+
+
+#if !defined(BOOST_MP_NO_CONSTEXPR_DETECTION) \
+  && !defined(DISABLE_TESTS) \
+  && defined(BOOST_MP_HAS_CONSTEXPR_DYNAMIC_ALLOC)
+
+// some operations fail with "Member call on variable whose lifetime has ended"
+// when using boost::multiprecision::et_on
```
mborland: Should we disallow et_on in the constant evaluation paths entirely so others don't hit this issue? The values have to be destroyed before run-time so there shouldn't be any runtime mixing of expression template options.

include/boost/multiprecision/cpp_int.hpp
marcoffee: Ok! Usually, I use braces for all blocks, but I try to follow the code convention, so if there are no braces in other one line blocks, I avoid adding them. There are other nearby blocks without braces, so I will add to them too

config/has_constexpr_dynamic_memory.cpp
marcoffee: Ok!

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: Trying to remember here, but I think it was me following the standard from other cases (e.g., [here](https://github.com/boostorg/multiprecision/blob/9e39fe57680e8d4fa91e5699a12f0c5015457113/include/boost/multiprecision/cpp_int/limits.hpp#L29)) or it was causing a compile time failure. I will try to compile without those checks and push if it works.

mborland: Prior to C++23 `constexpr` local variables could not also be static which is what those macro guards are for. In this case it looks like you are just using a `const` instead of a `constexpr` so it should work even with your C++20 requirement.

include/boost/multiprecision/cpp_int.hpp
marcoffee: I really don't know why I added those casts. I think they were useful when comparing objects that **might** have different type (e.g., [here](https://github.com/marcoffee/multiprecision/blob/dd3be65b5851db6b1dcf0987e28e61af597bb399/include/boost/multiprecision/cpp_int.hpp#L549)). I will try to compile without them (in all cases and only in cases where they are guaranteed to be equal) and push if it works.

include/boost/multiprecision/cpp_int/limits.hpp
```diff
@@ -24,33 +24,36 @@ inline BOOST_CXX14_CONSTEXPR_IF_DETECTION boost::multiprecision::number<boost::m
...
    constexpr
 #endif
-   const result_type                                                                                                                                                                          val = -result_type(~ui_type(0));
+   const result_type val = -result_type(~ui_type(0));
    return val;
 }
 
 template <std::size_t MinBits, std::size_t MaxBits, boost::multiprecision::cpp_integer_type SignType, boost::multiprecision::cpp_int_check_type Checked, class Allocator, boost::multiprecision::expression_template_option ExpressionTemplates>
-inline boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates>
+inline BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR_IF_DETECTION boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates>
```
jzmaddock: BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR_IF_DETECTION is a long name ;)

I assume it evaluates to just constexpr?  If so we can probably just call it BOOST_MP_CXX20_CONSTEXPR and be consistent with the other constexpr macros?

include/boost/multiprecision/cpp_int/limits.hpp
```diff
@@ -24,33 +24,36 @@ inline BOOST_CXX14_CONSTEXPR_IF_DETECTION boost::multiprecision::number<boost::m
...
-   // Bounded, signed, and an allocator (can't be constexpr).
-   using result_type = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates>                                               ;
+   // Bounded, signed, and an allocator (can be constexpr if allocations are constexpr).
+   using result_type = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates>;
    using ui_type = boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MaxBits, MaxBits, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked>, ExpressionTemplates>;
-   static const result_type                                                                                                                                                                          val = -result_type(~ui_type(0));
+#ifdef BOOST_MP_NO_CXX20_DYNAMIC_ALLOC_CONSTEXPR_DETECTION
+   static
+#endif
+   const result_type val = -result_type(~ui_type(0));
```
jzmaddock: This is something we should be careful of: in the C++20 case this code could be evaluated at either compile or run time.  In the latter case we recreate the local variable `val` each time it's run.  And given that val is a multiprecision type that's not actually a completely trivial thing.  So perhaps we could make `val` either `static const` or else `constexpr` depending on BOOST_MP_NO_CXX20_DYNAMIC_ALLOC_CONSTEXPR_DETECTION ?

include/boost/multiprecision/cpp_int/limits.hpp
```diff
@@ -61,103 +64,124 @@ get_min(const std::integral_constant<bool, true>&, const std::integral_constant<
...
+inline BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR_IF_DETECTION boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates>
 get_min(const std::integral_constant<bool, true>&, const std::integral_constant<bool, false>&, const std::integral_constant<bool, false>&)
 {
-   // Bounded and std::size_t with allocator (no constexpr):
-   static const boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates> val(0u);
+   // Bounded and std::size_t with allocator (can be constexpr if allocations are constexpr):
+#ifdef BOOST_MP_NO_CXX20_DYNAMIC_ALLOC_CONSTEXPR_DETECTION
+   static
+#endif
+   const boost::multiprecision::number<boost::multiprecision::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates> val(0u);
```
jzmaddock: Dynamic recreation of `val` each time the function is run again.

jzmaddock: I see there are more below, I'll let you work something out for all of them, rather than keep commenting ;)

include/boost/multiprecision/cpp_int/limits.hpp
jzmaddock: As noted in my comments above, we should find a better solution here, otherwise these variables get recreated/recalculated every time the function is run.  In other words making things compile for the constexpr case is dis-optimising the more common runtime case.

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -210,37 +296,45 @@ struct cpp_int_base<MinBits, MaxBits, signed_magnitude, Checked, Allocator, fals
...
-      constexpr data_type(limb_type i) noexcept : first(i) {}
-      constexpr data_type(signed_limb_type i) noexcept : first(static_cast<limb_type>(boost::multiprecision::detail::unsigned_abs(i))) {}
+      constexpr data_type() noexcept : la{0} {}
+      constexpr data_type(limb_type i) noexcept : la{i} {}
+      constexpr data_type(signed_limb_type i) noexcept : data_type(static_cast<limb_type>(boost::multiprecision::detail::unsigned_abs(i))) {}
 #if BOOST_MP_ENDIAN_LITTLE_BYTE
-      constexpr data_type(double_limb_type i) noexcept : double_first(i)
-      {}
-      constexpr data_type(signed_double_limb_type i) noexcept : double_first(static_cast<double_limb_type>(boost::multiprecision::detail::unsigned_abs(i))) {}
+      constexpr data_type(double_limb_type i) noexcept : la{}
```
jzmaddock: There are some changes in this file that make me distinctly nervous: this is a one such example, the initialization of data_type via a double_limb_type is a very deliberate choice that is hugely more efficient than using a loop.  I remember spending quite a bit of time getting these initializations as lean and fast as possible as they tend to be hit really rather often in many "real world" arithmetic where we need a few more bits here and there, but mostly only one or two limbs.  It basically makes the constructor so trivial as to not matter in this common case, where as as soon as there's a loop involved, most compilers won't even inline the code :(

eisenwave: Is that still a problem with modern compilers, or was that original designed based on less capable technology?

Also, nowadays you could also use `bit_cast` assuming C++20, which is implemented as an intrinsic in all compilers and should avoid any kind of loops or additional bitwise operations.

test/constexpr_test_dynamic_cpp_int.hpp
marcoffee: I agree. The expression parameters may contain a reference to another value, which breaks the validation done at constexpr evaluation time. Just a question, wouldn't those issues happen at runtime if, e.g., we return an expression from a function (using `auto` return) that uses local variables to build this expression?

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: I agree. This name is a mix between the `BOOST_MP_HAS_CONSTEXPR_DYNAMIC_ALLOC` and `BOOST_CXX14_CONSTEXPR_IF_DETECTION` (used on the other overloads). I added it in case some compiler just implements the dynamic allocation without the constexpr if detection, but I think it might not happen. In this case, is it ok if I just remove this macro and replace it with `BOOST_MP_HAS_CONSTEXPR_DYNAMIC_ALLOC`?

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: Unfortunately we will not be able to make them constexpr because no allocated heap storage can propagate from constexpr contexts to runtime. About making them static, I'm working on testing them

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: ;)

include/boost/multiprecision/cpp_int.hpp
marcoffee: I agree. The issue here is that constexpr tracks the current active member of an union and errors if we use the non-active one, so we need to initialize the one actively being used (in this case the `la` array). Also note that the loop runs in a constant number of iterations (`double_limb_type_limb_capacity` times, which is `static constexpr`), which means that the compiler will almost always turn this into a move operation or similar.

I may be able to use the `BOOST_MP_IS_CONST_EVALUATED(i)` check and initialize it in one shot when false or maybe use `std::bit_cast` (C++20 with `__cpp_lib_bit_cast` defined, might need another feature check) as suggested by @eisenwave

marcoffee:
> If you would add `LCOV_EXCL_LINE` or `LCOV_EXCL_START`/`LCOV_EXCL_STOP` to blocks that look like this:
> 
> ```c++
> if (BOOST_MP_IS_CONST_EVALUATED(n))
> {
>     // LCOV_EXCL_START
> 
>    ...
> 
>     // LCOV_EXCL_STOP
> }
> ```
> 
> Or `LCOV_EXCL_LINE` if it's a one liner.

There is just the case for [this](https://github.com/marcoffee/multiprecision/blob/dd3be65b5851db6b1dcf0987e28e61af597bb399/include/boost/multiprecision/cpp_int.hpp#L77) check that may run at consteval time or not depending on the input types. In this case, should I:
1) leave it without the LCOV_EXCL_*
2) add the LCOV_EXCL_*
3) split the check into two checks and add the LCOV_EXCL_* to the one with `BOOST_MP_IS_CONST_EVALUATED`?

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: I tried re-enabling those `static` modifiers and got the following error under ubuntu 24.04 running GCC 13:

```
| gcc.compile.c++ ../../../bin.v2/libs/multiprecision/test/test_numeric_limits_cpp_bin_float.test/gcc-13/debug/x86_64/threading-multi/visibility-hidden/test_numeric_limits.o
| In file included from ../../../boost/multiprecision/cpp_int.hpp:2442,
|                  from ../../../boost/multiprecision/cpp_bin_float.hpp:15,
|                  from test_numeric_limits.cpp:69:
| ../../../boost/multiprecision/cpp_int/limits.hpp: In function ‘constexpr boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, SignType, Checked, Allocator>, ExpressionTemplates> std::detail::get_min(const std::integral_constant<bool, true>&, const std::integral_constant<bool, true>&, const std::integral_constant<bool, false>&)’:
| ../../../boost/multiprecision/cpp_int/limits.hpp:45:29: error: ‘val’ defined ‘static’ in ‘constexpr’ function only available with ‘-std=c++2b’ or ‘-std=gnu++2b’
|    45 |    static const result_type val = -result_type(~ui_type(0));
|       |                             ^~~
```

I could try only removing those `static` on consteval contexts (which would ensure the build will only happen during constexpr evaluation or runtime startup), but it might either require `if consteval` (from C++23) or `BOOST_MP_IS_CONST_EVALUATED` (with some trick to make the compiler think the variable we pass is not constant evaluated) and some code repetition/macros to initialize `val` in both contexts. I could also try to move those variables to the outer scope and just return them, which might work. I will try the last option and report it here.

marcoffee: Unfortunately the last option did not work, as I cannot return a non-constexpr static variable (they were marked static const) from a constexpr evaluated function. Adding constexpr to the variable would also not work because heap allocations must stay in constexpr evaluation, i.e., they cannot "leak" to runtime.

During the tests I think I found another solution. Since types with an allocator can be initialized from types without one, I can just call the non-allocator method and just construct the resulting type. I will try it and report the results as before.

include/boost/multiprecision/cpp_int/limits.hpp
mborland: Does Boost.Config have `BOOST_CXX20_CONSTEXPR`? I guess I would associate that with `__cpp_constexpr > 201907L` since there was a huge expansion of constexpr in 20. I would go with `BOOST_MP_CXX20_ALLOC_CONSTEXPR`. A macro with `_HAS_` in the name looks odd in the function signature, usually those are associated with some in body configuration or configuration of other macros.

test/constexpr_test_dynamic_cpp_int.hpp
mborland: Yes. We happen to have this warning on the front page of the documentation:

image

mborland:
> 
> There is just the case for [this](https://github.com/marcoffee/multiprecision/blob/dd3be65b5851db6b1dcf0987e28e61af597bb399/include/boost/multiprecision/cpp_int.hpp#L77) check that may run at consteval time or not depending on the input types. In this case, should I:
> 
> 1. leave it without the LCOV_EXCL_*
> 2. add the LCOV_EXCL_*
> 3. split the check into two checks and add the LCOV_EXCL_* to the one with `BOOST_MP_IS_CONST_EVALUATED`?

In that case it looks like you could write tests that would cover both paths at runtime. One with same DestT and SrcT and one where they are different.

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: > During the tests I think I found another solution. Since types with an allocator can be initialized from types without one, I can just call the non-allocator method and just construct the resulting type. I will try it and report the results as before.

This solution worked. Regarding @jzmaddock comments about performance, I just found out that on `return 0` cases the compilers (g++-13 and clang++-19) return a much more efficient assembly for non-static cases (compiler explorer comparison: [non-static](https://godbolt.org/z/q5sdcaTz5) vs [static](https://godbolt.org/z/hPKjs1YoE)). This is not the case for more complex returns, such as minimum of signed integers (compiler explorer comparison: [non-static](https://godbolt.org/z/78971q5d1) vs [static](https://godbolt.org/z/TzE3rrKbb)).

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: I will use the `BOOST_MP_CXX20_DYNAMIC_ALLOC_CONSTEXPR` then. I prefer using `DYNAMIC_ALLOC` instead of just `ALLOC` because it closer to the standard feature tests `__cpp_constexpr_dynamic_alloc` and `__cpp_lib_constexpr_dynamic_alloc`, but I can change it to `BOOST_MP_CXX20_ALLOC_CONSTEXPR` if required.

include/boost/multiprecision/cpp_int.hpp
marcoffee: @jzmaddock after some code reading I think I found a good solution for this. I employed the idea from the other union constructors (to initialize using the `double_first` then, if in constexpr context, call another constructor that receive the limbs separately (e.g., [here](https://github.com/boostorg/multiprecision/blob/9e39fe57680e8d4fa91e5699a12f0c5015457113/include/boost/multiprecision/cpp_int.hpp#L583)). It looks like this strategy passes the constexpr tests.

I just have some questions about the code:

1) Only this `data_type` union disallows construction using `double_limb_type` when the CPU is not little-endian ([here](https://github.com/boostorg/multiprecision/blob/9e39fe57680e8d4fa91e5699a12f0c5015457113/include/boost/multiprecision/cpp_int.hpp#L231)), while the others allow it ([here](https://github.com/boostorg/multiprecision/blob/9e39fe57680e8d4fa91e5699a12f0c5015457113/include/boost/multiprecision/cpp_int.hpp#L583), and [here](https://github.com/boostorg/multiprecision/blob/9e39fe57680e8d4fa91e5699a12f0c5015457113/include/boost/multiprecision/cpp_int.hpp#L776)). Should I add this check to all of those constructors or there is a reason for this?

2) It seems like the `m_first_limb`fields from the other unions are unused. Can I just remove them along this commit?

Thanks for your review :)

marcoffee:
> In that case it looks like you could write tests that would cover both paths at runtime. One with same DestT and SrcT and one where they are different.

I found out that this method is only called for copying limbs, so the types should always be the same. I just removed the possibility of passing two different types and ended up with just the `BOOST_MP_IS_CONST_EVALUATED`, where I added the `LCOV_EXCL_*` comments.

test/constexpr_test_dynamic_cpp_int.hpp
marcoffee: I just figured out what was the issue with expression templates. It was [this](https://github.com/marcoffee/multiprecision/blob/dd3be65b5851db6b1dcf0987e28e61af597bb399/test/constexpr_test_dynamic_cpp_int.hpp#L343) test that uses a lambda to generate a value and return it. Since [lambdas auto detect the output type](https://en.cppreference.com/w/cpp/language/lambda.html#cite_ref-1), it was returning an `expression`. All the other tests work with `et_on`, so I will enable them and fix this case.

With that in mind, I think it would be best if we added another warning about returning expression templates from lambdas that do not have an implicit return type.

Thanks for the heads up

EDIT

I found out that clang refuses to compile the integral functions tests when expression templates are enabled. I tested up to clang++-21 and all of them fail indicating that there is a temporary being accessed. 

Note that those methods do not work in constexpr even for the current published version. If I just disable the expression templates they work fine. Tests (on compiler explorer): [clang++-16 to clang++-21 with `et_on`](https://godbolt.org/z/fn69aGsWs) vs [clang++-16 to clang++-21 with `et_off`](https://godbolt.org/z/djvM9zs1z).  This is the same error that I'm getting with the updated code here. I can create an issue later (if there is not a previous one about that).

In this case, I was thinking of disabling just those tests when compiling via clang and mention the issue on a comment next to the `BOOST_CLANG` check.

marcoffee: > I found out that clang refuses to compile the integral functions tests when expression templates are enabled. I tested up to clang++-21 and all of them fail indicating that there is a temporary being accessed.

@mborland I found a simple solution for this issue, but since it is affecting the current published version, I would like to know if I should open an issue (and a PR) or just another PR to fix this.

mborland: If you would, please add that fix in a separate PR and then rebase this one on top of develop once it is merged.

marcoffee: Ok, thanks for the reply.

Created #757

include/boost/multiprecision/cpp_int.hpp
```diff
@@ -63,6 +64,91 @@ namespace detail {
...
+   template <class DestT, class SrcT>
+   static BOOST_MP_CXX14_CONSTEXPR DestT* constexpr_copy(DestT* dest, const SrcT* src, std::size_t n)
+   {
+      bool use_loop = !std::is_same<DestT, SrcT>::value;
+
+#     ifndef BOOST_MP_NO_CONSTEXPR_DETECTION
+      use_loop = use_loop || BOOST_MP_IS_CONST_EVALUATED(n);
+#     endif
+
+      if (use_loop)
```
jzmaddock: This (and related functions below) also make me nervous, here we have a non-constant runtime variable being used in a conditional followed by a loop.  From memory, either of those will tend to rule out inline expansion.

In comparison, the functions memcpy/memset/memmove are so heavy optimised, and indeed so heavily used that folks even design their chips around them!

At the very least we need to make the conditional an `if constexpr` (which we can as all the conditions are known at compile time), and also verify that the new code is as lean as the old and results in direct inline/intrinsic expansion of memcpy etc without any additional function calls or runtime code.

jzmaddock: Oh, and the function needs to be declared inline as well ;)

marcoffee: @jzmaddock thanks for the comment.

In this case, the use was only used when `SrcT` is not the same as `DestT`, which disallows usage of `memcpy`/`memmove` because of alignment. Anyways, this code has changed because [I removed the possibility of `SrcT` being different from `DestT`](https://github.com/boostorg/multiprecision/pull/654#issuecomment-4220606123), so we only use the loop during constexpr evaluation now.

About the inlining, I will replace `BOOST_MP_CXX14_CONSTEXPR` with a `#if`/`#endif` check.

I'm just waiting for the reply on the other comments to push the updated code, but I can push it now if needed.

marcoffee:
Just resolved the conflicts locally and started the tests. As soon as it is finished here I will push a new version with the changes.

marcoffee:
It seems like the Install Packages step is failing due to network errors. I'm trying to force push to the branch to trigger rerun.

marcoffee:
> 1 Missing ⚠️

I could not cover [this line](https://app.codecov.io/gh/boostorg/multiprecision/pull/654/blob/include/boost/multiprecision/cpp_int.hpp#L1908). Since the previous and next line are covered, I'm assuming the compiler is optimizing out this assignment because it can prove that `val` is always reassigned after. I cannot remove the `= 0` (probably the reason `codecov` is assuming it is a valid line, different from [the previous coverage](https://app.codecov.io/gh/boostorg/multiprecision/pull/757/blob/include/boost/multiprecision/cpp_int.hpp?dropdown=coverage#L1793)), otherwise it will not compile for standards less than C++20 due to uninitialized variables (even if that variable is always initialized later, which is the case here).

I can add `LCOV_EXCL_LINE`, but this line is valid for runtime contexts when the compiler is unable to optimize it out (e.g., [similar case](https://app.codecov.io/gh/boostorg/multiprecision/pull/654/blob/include/boost/multiprecision/cpp_int.hpp#L1949)).

ckormanyos:
> I could not cover

Line 1908 unconditionally follows line 1907 (where 1907 is covered).

This is, in my opinion, a clear false-negative. You can `LCOV_EXCL_LINE` on 1908.

Coverage is challenging with optimized C++. Sometimes it's just not right in LCOV. In some other projects, we specifically reduce the optimization level on coverage runs. This is also not perfect. In the Multiprecision case, we collect coverage from relatively aggressively optimized runs. And it takes a keen eye to find false negatives.

Good catch @marcoffee

ckormanyos:
I don't think you need to mandate 100% coverage on the patch. Other opinions will vary.

But the entire project is around 96% right now. So we are looking for , maybe, close to 100% patch coverage. The file you mention goes on to have entirely uncovered blocks. These might be more interesting. Or they simply might be out of the scope of your changes.

marcoffee:
> Good catch @marcoffee

@ckormanyos Thanks!

> Or they simply might be out of the scope of your changes.

Those other blocks were already uncovered before the PR. I can try to cover them now, but it will add more information to review on the PR and I agree that it might be out of scope. Another option is opening a new PR after this one is merged to try to cover those lines, but the new one would focus entirely on tests (and some small refactoring). 

Curiously enough [this very similar line](https://app.codecov.io/gh/boostorg/multiprecision/pull/654/blob/include/boost/multiprecision/cpp_int.hpp#L1949) is marked as covered.

ckormanyos:
Marco (@marcoffee) I would recommend that you do NOT try to cover previously uncovered blocks within the scope of your PR.

This is something we will work on in our team as we evolve.

You concentrate on your scope of changes.

include/boost/multiprecision/cpp_int/limits.hpp
mborland: @marcoffee were you able to make these work as `static const` or something similar so we don't have to recreate the variable each time?

mborland:
I just took a look through this again and it all looks good to me. One thing I think would be worth doing is running some of our benchmarks for `cpp_int` such as in performance/test_performance.cpp at develop and at the head of your PR to make sure that changes required for `constexpr` support didn't cause any runtime regressions such as in a few places I think John pointed out.

include/boost/multiprecision/cpp_int/limits.hpp
marcoffee: @mborland for the cases returning `0` I found out that simply building the value every time generates a better assembly than using `static` (comment with compiler explorer proof [here](https://github.com/boostorg/multiprecision/pull/654#discussion_r3059550675), code [here](https://github.com/marcoffee/multiprecision/blob/d0b80d2bd8a31a63e8871d5af618f8302fbad587/include/boost/multiprecision/cpp_int/limits.hpp#L91)). For the other cases where `static const` would not work due to `constexpr` evaluation limitations I'm using the `static const`/`static constexpr` values from the overload that accepts static and copying to the result, which would be the case of `static const` because since we are not returning a reference, the value would still be copied at every invocation (comment [here](https://github.com/boostorg/multiprecision/pull/654#discussion_r3055787596), code [here](https://github.com/marcoffee/multiprecision/blob/d0b80d2bd8a31a63e8871d5af618f8302fbad587/include/boost/multiprecision/cpp_int/limits.hpp#L67)).

PR #655, opened 2025-02-20
Test constexpr initialization

ohhmm:


jzmaddock:
That test as it stands is not going to work, we also have separate constexpr tests for cpp_int in https://github.com/boostorg/multiprecision/blob/develop/test/constexpr_test_cpp_int.cpp and related test files which I believe cover this.

What was it you thought was missing?

PR #657, opened 2025-02-24
fix the mpfr stack str bug of the deadloop in the `while`

FAKERINHEART:


ckormanyos:
Oh Also, @FAKERINHEART, dou you have a test case that exercises this change to the `for`-loop?. That would be nice because then we could add the test case as well for the new lines.

jzmaddock:
Actually... is there any reason not to just delete the whole outer loop?  If we enter the loop then the condition that leads to the break will always be taken?

ckormanyos:
> Actually... is there any reason not to just delete the whole outer loop? If we enter the loop then the condition that leads to the break will always be taken?

Do you mean to remove the outer `while`-loop at line $848$?

I agree @jzmaddock.

- But then also remove the `break;` statement at line $853$ that, at the moment, breaks from the outer `while`-loop, right?
- Do we need a range-check on `offset + 1`?

Cc: @FAKERINHEART

FAKERINHEART:
> Oh Also, @FAKERINHEART, do you have a test case that exercises this change to the `for`-loop?. That would be nice because then we could add the test case as well for the new lines.

No, I think that it may has no needs to add a new test. Please look into mpfr.hpp:431, and its commit log: 485fa7c9ae5c568475b376bb5bbf6d123003901e. It was fixed for the `mpfr_float_imp<digits10, allocate_dynamic>`, but it forgot to fix `mpfr_float_imp<digits10, allocate_stack>` which is this merge case. They are the same condition, and it has been add tests in the 485fa7c9ae5c568475b376bb5bbf6d123003901e.

FAKERINHEART:
CI has been passed.
Could we merge this PR please?
@ckormanyos

jzmaddock:
Bare with me, I'm not convinced this is the correct fix (or indeed that there is an issue, as the whole loop appears to be a mistake that can never actually loop anyway), I need to investigate some more....

jzmaddock:
My apologies: you're quite correct, I was completely misreading the code until I saw it with highlighting in my editor!  My bad, will merge...

mborland:
I'll merge this into master for 1.88 since we're still safe to do so

FAKERINHEART:
thx~

PR #663, opened 2025-03-17
Don't use deprecated form of user-defined literal operator

Thomas-Barbier-1A:


ckormanyos:
Thanks @Thomas-Barbier-1A for catching this and @mborland for the quick fix for the next release!

PR #667, opened 2025-03-24
Fix compilation when building with msvc's new preprocessor

pps83:
fixes the following error when standard conforming preprocessor is enabled (`/Zc:preprocessor` option)
```
1>time.cpp
1>D:\work-pps\boost_1_87_0\boost\multiprecision\cpp_int\literals.hpp(270,1): warning C5103: pasting '""_cppi' and '128' does not result in a valid preprocessing token
1>(compiling source file '../src/time.cpp')
1>    D:\work-pps\boost_1_87_0\boost\multiprecision\cpp_int\literals.hpp(252,9):
1>    in expansion of macro 'BOOST_MP_DEFINE_SIZED_CPP_INT_LITERAL'
1>    D:\work-pps\boost_1_87_0\boost\config\helper_macros.hpp(33,9):
1>    in expansion of macro 'BOOST_JOIN'
1>    D:\work-pps\boost_1_87_0\boost\config\helper_macros.hpp(34,9):
1>    in expansion of macro 'BOOST_DO_JOIN'
1>    D:\work-pps\boost_1_87_0\boost\config\helper_macros.hpp(35,9):
1>    in expansion of macro 'BOOST_DO_JOIN2'
1>D:\work-pps\boost_1_87_0\boost\multiprecision\cpp_int\literals.hpp(270,1): error C2988: unrecognizable template declaration/definition
```

### DIFF
--- a/include/boost/multiprecision/cpp_int/literals.hpp
+++ b/include/boost/multiprecision/cpp_int/literals.hpp
@@ -249,29 +249,30 @@ constexpr typename boost::multiprecision::literals::detail::unsigned_cpp_int_lit
    return boost::multiprecision::literals::detail::make_backend_from_pack<pt, typename boost::multiprecision::literals::detail::unsigned_cpp_int_literal_result_type<static_cast<unsigned>((sizeof...(STR)) - 2u)>::backend_type>::value;
 }
 
-#define BOOST_MP_DEFINE_SIZED_CPP_INT_LITERAL(Bits)                                                                                                                                                                                \
-   template <char... STR>                                                                                                                                                                                                          \
-   constexpr boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<Bits, Bits, boost::multiprecision::signed_magnitude, boost::multiprecision::unchecked, void> > operator BOOST_JOIN(""_cppi, Bits)()    \
-   {                                                                                                                                                                                                                               \
-      using pt = typename boost::multiprecision::literals::detail::make_packed_value_from_str<STR...>::type;                                                                                                                       \
-      return boost::multiprecision::literals::detail::make_backend_from_pack<                                                                                                                                                      \
-          pt,                                                                                                                                                                                                                      \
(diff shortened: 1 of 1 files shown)


mborland:
> Perhaps, it's better to
> 
> ```c++
> #define BOOST_MP_LIT(P, N) BOOST_JOIN(operator"", BOOST_JOIN(P, N))
> ```
> 
> This way grepping for `operator""` would work. Lmk if I should update the PR

I think this would be better

pps83:
> ...
> This way grepping for `operator""` would work. Lmk if I should update the PR

applied. This way it's better

mborland:
> > ...
> > This way grepping for `operator""` would work. Lmk if I should update the PR
> 
> applied. This way it's better

Thanks. I approved the CI run

jzmaddock:
How does this interact with https://github.com/boostorg/multiprecision/pull/665 and https://github.com/boostorg/multiprecision/pull/663 ?

mborland:
> How does this interact with #665 and #663 ?

Good catch. I just checked out this commit and it does not cause "-Wdeprecated-literal-operator" to be emitted. I believe it's safe since that was the only intention of those two PRs.

pps83:
> How does this interact with #665 and #663 ?

I also build with clang and had these warnings. After I took latest updates that fix the warnings I had issues with msvc. With this PR both issues should be resolved now.

pps83:
let me know if I should create a new pr and remove `BOOST_MP_LIT` thing completely

mborland:
I think it's fine. We can still grep for operator "" in the changes from this PR which we weren't able to do previously.

pps83:
I didn't think about it initially, but after looking into it, I don't see a reason for using `BOOST_JOIN` there. I think it's better to remove it, so copy pasta doesn't get replicated. join would be needed if `BOOST_MP_DEFINE_SIZED_CPP_INT_LITERAL` wasn't used with actual literals.

pps83:
> I think it's fine. We can still grep for operator "" in the changes from this PR which we weren't able to do previously.

I added update: https://github.com/boostorg/multiprecision/pull/669 ignore/close it if you think it's unnecessary

Issue #673, opened 2025-04-08
Hard-to-find warning implicit-fallthrough with NO_EXCEPTIONS


ckormanyos:
Found an interesting, yet hard-to-find, warning today in modular boost multiprecision.

When I compile code with `BOOST_NO_EXCEPTIONS`, there is a line in `cpp_bin_float` that expands differently in this compiler-configuration. And the warning `-Wimplicit-fallthrough` is issued [here](https://github.com/boostorg/multiprecision/blob/02d8862203d0cd95b81c51d0dd30cdffad1965c9/include/boost/multiprecision/cpp_bin_float.hpp#L1617).

The implicit fallthrough is at line 1616. When `BOOST_MP_THROW_EXCEPTION` does not actually throw, the compiler correctly detects the implicit fallthrough and warns on `-Wextra`.

So we can just add the same two lines that the case below has, since this is the behavior that was being fallen through to.

ckormanyos:
I can handle this one, but I'll ask for review when it cycles.

Cc: @jzmaddock

PR #674, opened 2025-04-08
Handle implicit fallthrough warn on no-except

ckormanyos:


ckormanyos:
See #673

ckormanyos:
> I guess it's time to upgrade the CI system too. Ubuntu Focal will be removed by Github next week

Fair enough. Thanks @mborland. What do you think I should do with all those tests? Should I puth them on ubuntu-latest. or 22.04? Should we simply drop them?

Cc: @jzmaddock

jzmaddock:
The fix is fine, but I have a hunch will generate "unreachable code" warnings when building with exceptions ON.  Maybe put those lines inside a #if block?  Annoying I know....

ckormanyos:
> The fix is fine, but I have a hunch will generate "unreachable code" warnings when building with exceptions ON. 

Yep, goti it. I will be super-sure to decorate with the PP so that no corollary warnings arise. Nich point John (@jzmaddock).

Will hit this after the CI blitz subsides...

ckormanyos:
Drone is way too long.  Whatever on earth is running there, we need to reduce the scope of testing on that thing. I'm looking for half an hour or less. This multi-hour thing isn't how it's supposed to work.

I'll try to figure out what's taking forever there. and reduce the cycles. But that'll be after this PR.

Cc: @mborland and @jzmaddock

mborland:
It looks like you're waiting on s390x runners. They are in pretty limited supply, but generally available. 26 mins is the longest single runner that's complete. Let me ping the guy that manages drone.

ckormanyos:
> It looks like you're waiting on s390x runners. They are in pretty limited supply, but generally available. 26 mins is the longest single runner that's complete. Let me ping the guy that manages drone.

Ah thanks Matt (@mborland) I hadn't thought of that exactly. Are you waiting for run-time to trickle through, ...
Or are you wainting for a runner to even start.

Either way, it's a lag-time that quite honestly needs to be critically examined and weighed regarding added value.

mborland:
Waiting for the runner to even start. I'm willing to bet it's related to release going out today since every boost library is showing recently updated. Basically DDoSed drone.

ckormanyos:
> Waiting for the runner to even start. I'm willing to bet it's related to release going out today since every boost library is showing recently updated. Basically DDoSed drone.

10-4.
I will chill and check it out.

PR #675, opened 2025-04-08
Update github actions

mborland:
We got hit with the brownout. I don't see a good way to go wholesale with boost.ci since we have so many test sets we need to cut down. I'll update drone next since the most recent clang on there is 10 and version 20 is on the streets.

CC: @ckormanyos

### DIFF
--- a/.github/workflows/multiprecision.yml
+++ b/.github/workflows/multiprecision.yml
@@ -144,14 +144,14 @@ jobs:
         run: cat bin.v2/config.log
         working-directory: ../boost-root/
   ubuntu-focal:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     defaults:
       run:
         shell: bash
     strategy:
       fail-fast: false
       matrix:
-        compiler: [ g++-10, g++-11, clang++-10, clang++-11 ]
+        compiler: [ g++-13, g++-14, clang++-19, clang++-20 ]
         standard: [ c++14, c++17 ]
         suite: [ github_ci_block_1, github_ci_block_2 ]
     steps:
@@ -175,7 +175,7 @@ jobs:
         if: steps.retry1.outcome=='failure'
         run: sudo apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
       - name: Install packages
-        run: sudo apt install g++-9 g++-10 g++-11 clang-10 clang-11 libgmp-dev libmpfr-dev libtommath-dev libeigen3-dev libmpfi-dev libmpc-dev
+        run: sudo apt install g++-13 g++-14 clang-19 clang-20 libgmp-dev libmpfr-dev libtommath-dev libeigen3-dev libmpfi-dev libmpc-dev
       - name: Checkout main boost
         run: git clone -b develop --depth 1 https://github.com/boostorg/boost.git ../boost-root
       - name: Update tools/boostdep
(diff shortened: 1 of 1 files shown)


mborland:
MSVC 14.3 with C++20 is failing on `git_issue_509.cpp(15): error C2039: 'bit_cast': is not a member of 'std'`

And the standalone test are failing with this:

```
standalone_test_convert_from_tom_int.cpp: In instantiation of ‘void test_convert_to_builtin() [with From = boost::multiprecision::number<boost::multiprecision::backends::tommath_int>; To = float]’:
standalone_test_convert_from_tom_int.cpp:199:43:   required from here
standalone_test_convert_from_tom_int.cpp:182:15: error: cannot convert ‘boost::multiprecision::number<boost::multiprecision::backends::tommath_int>’ to ‘float’ in initialization
  182 |       To t1 = from;
      |               ^~~~
      |               |
      |               boost::multiprecision::number<boost::multiprecision::backends::tommath_int>

    "g++-13"   -fvisibility-inlines-hidden -std=c++14 -fPIC -m64 -pthread -O0 -fno-inline -Wall -g -fvisibility=hidden -Wall -Wextra -DBOOST_ALL_NO_LIB=1 -DCI_SUPPRESS_KNOWN_ISSUES -DHAS_GMP -DHAS_MPFR -DHAS_TOMMATH   -I"../../.." -I"../include"  -c -o "../../../bin.v2/libs/multiprecision/test/standalone_test_convert_from_tom_int.test/gcc-13/debug/x86_64/threading-multi/visibility-hidden/standalone_test_convert_from_tom_int.o" "standalone_test_convert_from_tom_int.cpp"
```

ckormanyos:
> Looks like the new failures are all in example/gauss_laguerre_quadrature.cpp which I think was @ckormanyos ?

Yeah, that's the guy. I just pushed a presumed fix.

But @jzmaddock I think there might be another conversion failure lurking around, but my glance was super quick.

Let's see if I dialed the quadrature first then if that's green if there is anything else.

Cc: @mborland

mborland:
> But @jzmaddock I think there might be another conversion failure lurking around, but my glance was super quick.
> 

Yep. Conversion in standalone is above. I just pushed the fix for the MSVC 14.3 issue

ckormanyos:
Hi @mborland and @jzmaddock this is a good one and it is nice to move forward.

It looks like there still might be some kind of a bogey in `standalone_test_convert_from_tom_int` on some compilers. I have not isolated it yet, but I'm not sure if I'm the right guy for that phenomenon.

mborland:
It looks like the updated ubuntu version brings a new version of Tommath, which had not been updated since 2019. Maybe some of the deprecated things are now gone? Might take a minute to peel this one apart

jzmaddock:
It's a strange one for sure.... but the error is in the front end so the tommath version should *probably* not matter?

jzmaddock:
The error can be reproduced with msvc both with and without the #define BOOST_MP_STANDALONE.  I have a hunch this test wasn't being run before?

jzmaddock:
I have a fix: there was a copy-and-paste error when the file was created, will push a fix shortly...

mborland:
Thanks @jzmaddock! I guess similar updates should be applied against the math CI system too. I'll open a PR.

mborland:
CI is back to green now. Thanks for the help!

PR #676, opened 2025-04-15
Uniform logic has or not-has GCD/LCM

ckormanyos:
In [cpp_int/misc.hpp](https://github.com/boostorg/multiprecision/blob/ce635570f667266864c0f18d1d9accd7b591486a/include/boost/multiprecision/cpp_int/misc.hpp#L29), there is a dependency on `Boost.Integer`.

But if and only if it is not standalone. Yet the standalone version does not need integer. I was wondering if it is possible to remove this dependency in the non-standalone mode? Thereby the behavior would be the same for `constexpr_gcd` in standalone as well as non-standalone mode.

Why? I got some pesky warnings coming in from `Boost.Integer`.

The header has also been refactored to treat having/not-having `std::gcd` in a uniform fashion with respect to compiler switches. The compiler switches were not the same in the two places earlier.

Cc: @jzmaddock and @mborland

### DIFF
--- a/include/boost/multiprecision/cpp_int/misc.hpp
+++ b/include/boost/multiprecision/cpp_int/misc.hpp
@@ -11,29 +11,32 @@
 #ifndef BOOST_MP_CPP_INT_MISC_HPP
 #define BOOST_MP_CPP_INT_MISC_HPP
 
-#include <boost/multiprecision/detail/standalone_config.hpp>
-#include <boost/multiprecision/detail/number_base.hpp>
 #include <boost/multiprecision/cpp_int/cpp_int_config.hpp>
-#include <boost/multiprecision/detail/float128_functions.hpp>
 #include <boost/multiprecision/detail/assert.hpp>
-#include <boost/multiprecision/detail/constexpr.hpp>
 #include <boost/multiprecision/detail/bitscan.hpp> // lsb etc
+#include <boost/multiprecision/detail/constexpr.hpp>
+#include <boost/multiprecision/detail/float128_functions.hpp>
 #include <boost/multiprecision/detail/hash.hpp>
 #include <boost/multiprecision/detail/no_exceptions_support.hpp>
+#include <boost/multiprecision/detail/number_base.hpp>
+#include <boost/multiprecision/detail/standalone_config.hpp>
+
+#ifndef BOOST_MP_STANDALONE
+#include <boost/integer/common_factor_rt.hpp>
+#endif
 
 #ifdef BOOST_MP_MATH_AVAILABLE
 #include <boost/math/special_functions/next.hpp>
 #endif
 
-#include <type_traits>
-#include <stdexcept>
-#include <cmath>
-
 #if (defined(__cpp_lib_gcd_lcm) && (__cpp_lib_gcd_lcm >= 201606L)) && (!defined(BOOST_HAS_INT128) || !defined(__STRICT_ANSI__))
-#include <numeric> // std::gcd
-
-#define BOOST_MP_STD_GCD_AVAILABLE
+#define BOOST_MP_LIB_GCD_LCM_AVAILABLE
 #endif
+
+#include <cmath>
+#include <numeric>
+#include <stdexcept>
+#include <type_traits>
 
 #ifdef BOOST_MSVC
 #pragma warning(push)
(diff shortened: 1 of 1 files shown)


include/boost/multiprecision/cpp_int/misc.hpp
```diff
@@ -20,17 +20,19 @@
...
+#endif
+
 #include <type_traits>
 #include <stdexcept>
 #include <cmath>
 
-#ifndef BOOST_MP_STANDALONE
-#include <boost/integer/common_factor_rt.hpp>
-#endif
+#if (defined(__cpp_lib_gcd_lcm) && (__cpp_lib_gcd_lcm >= 201606L)) && (!defined(BOOST_HAS_INT128) || !defined(__STRICT_ANSI__))
```
jzmaddock, line 32: I wonder if we really need the `&& (!defined(BOOST_HAS_INT128) || !defined(__STRICT_ANSI__))` part?  Ah wait... no std::gcd and std::lcm for __int128?  I know we have our own gcd/lcm as a fall-back of last resort, but they're not actually very performant (I suspect anyway) compared to the full Boost versions.  Or perhaps I'm wrong on that?

In any case we need to get these CI tests fixed up before we can do much of anything... :(

ckormanyos:
> ... we have our own gcd/lcm as a fall-back of last resort, but they're not actually very performant (I suspect anyway) compared to the full Boost versions. Or perhaps I'm wrong on that?

Hmmm... That's a good one. I would be somewhat leary of incurring a performance loss for this. The real changes I implemented part involve calculating the `gcd()` for integral types. These default to `::std::gcd()` (when available).

I could work out some perf-tests before/after.

> In any case we need to get these CI tests fixed up before we can do much of anything... 

Well, it _can_ go green and just did.

We can also drop this one any time, it is not a critical change. But I'll invest the hour of play-time for some perf-experiments...

Thanks John!

ckormanyos:
>> ... we have our own gcd/lcm as a fall-back of last resort, but they're not actually very performant (I suspect anyway) compared to the full Boost versions. Or perhaps I'm wrong on that?

> work out some perf-tests before/after.

Well, the performance numbers took a while, but here they are. Care was taken to measure the time of the GCD operations themselves insofar as possible (maybe with extra time only for `move`-ingh them into the vector storage.

So it does, in fact, appear that the full Boost versions of GCD in `Boost.Integer` offer slightly superior performance. In light of this, I don't see any reason to pursue this PR any further. And I'll close it if that's the consensus?

Cc: @jzmaddock and @mborland 

|     digits    |    runs           |    time (develop) [s]  |    time (candidate) [s] |    ratio     |
|---------------|-------------------|------------------------|-------------------------|--------------|
|     128       |  16 * 0x200000    |      1.65              |       1.63              |       1.00   |
|     256       |  16 * 0x200000    |      2.20              |       2.27              |       1.03   |
|     256       |  16 * 0x200000    |      3.48              |       3.64              |       1.05   |
|    1024       |  16 * 0x200000    |      6.29              |       6.75              |       1.07   |
|    4096       |  16 * 0x200000    |      23.3              |      24.6               |       1.06   |

ckormanyos:
> The header has also been refactored to treat having/not-having `std:.gcd` in a uniform fashion with respect to compiler switches.

This part, however, might be a useful change.

Cc: @jzmaddock and @mborland

ckormanyos:
Add the source codes used for performance measurements. It is a zip archive of the source files.
[Test.zip](https://github.com/user-attachments/files/20037410/Test.zip)

ckormanyos:
Hi @jzmaddock and @mborland I've never really follwed the Circle CI build that much.

For some reason, on trivial changes, the `inspect` job is failing. But I can't really see why. Do you guys know why it is failing?

jzmaddock:
> For some reason, on trivial changes, the inspect job is failing. But I can't really see why. Do you guys know why it is failing?

No, but it's started failing on Math as well.  I think something outside our control has changed, will try to take a look...

jzmaddock:
See comment: https://github.com/boostorg/inspect/commit/0ef79484318eae34922500f0ad8b72806e33ac78#r156401450

ckormanyos:
Merging in acceptance of upstream break in `Inspect`.

Issue #678, opened 2025-04-24
Namespace in `std::variant` header collidies with `boost::multiprecision` serialization namespace


jeffro256:
Cross-posting https://github.com/boostorg/serialization/issues/328 since a modification in either repository is acceptable.

PR #680, opened 2025-04-24
Remove namespace alias from not our namespace

mborland:
Closes: #678


jeffro256:
Looks good, thank you!

PR #683, opened 2025-04-26
A few header-only warns and warns in CI

ckormanyos:
In this PR, we have found a few remaining warnings `-Wconversion` and `-Wsign-conversion` in the code and cleaned these up.

There was also a deprecated test header being used in the setup of the specfun tests. This has been corrected and the output logs of the specfun tests are expected to be much cleaner.


jzmaddock:
Thanks Chris!

Issue #652, opened 2025-01-20
Possible error convert bin_float value bigger than unsigned_long_long_max to unsigned long long


ckormanyos:
OK so,

- You take a number larger than `::std::numeric_limits<unsigned long long>::max()` and store this in `cpp_bin_float_50`.
- Then you convert that `cpp_bin_float_50` value to `unsigned long long`
- I find the wrong answers, or what seem to be wrong. I expect to get the `max()`- value of the integral type.
- `cpp_dec_float_50` gets it right, if that is what right means.
- Both backends behave properly for `signed long long`, the signed version of the intger.

```cpp

#include <boost/multiprecision/cpp_bin_float.hpp>
#include <boost/multiprecision/cpp_dec_float.hpp>

#include <ctime>
#include <iomanip>
#include <iostream>
#include <limits>
#include <random>

template<typename FloatType, typename LongIshType>
auto test() -> void
{
  std::mt19937_64 gen { static_cast<typename std::mt19937_64::result_type>(std::clock()) };

  auto dis =
    std::uniform_real_distribution<float>
    {
      static_cast<float>(1.01F),
      static_cast<float>(1.04F)
    };

  for(auto i = static_cast<unsigned>(UINT8_C(0)); i < static_cast<unsigned>(UINT8_C(8)); ++i)
  {
    static_cast<void>(i);

    using float_type = FloatType;

    using long_type = LongIshType;

    float_type
      flt_max_more
      {
          float_type { (std::numeric_limits<long_type>::max)() }
        * static_cast<float>((static_cast<int>(i) + 1) * 2) * dis(gen)
      };

    const long_type conversion_result_max_more { static_cast<long_type>(flt_max_more) };

    std::stringstream strm { };

    strm << "0x" << std::hex << std::uppercase << conversion_result_max_more;

    std::cout << strm.str() << std::endl;
  }

  std::cout << std::endl;
}

auto main() -> int
{
  test<::boost::multiprecision::cpp_dec_float_50, signed long long>();
  test<::boost::multiprecision::cpp_bin_float_50, signed long long>();

  test<::boost::multiprecision::cpp_dec_float_50, unsigned long long>();
  test<::boost::multiprecision::cpp_bin_float_50, unsigned long long>();
}
```

I get (the last group is seemingly wrong):

```sh
0x7FFFFFFFFFFFFFFF
0x7FFFFFFFFFFFFFFF
0x7FFFFFFFFFFFFFFF
[...]

PR #684, opened 2025-04-29
Handle issue652 with upper-bound check

ckormanyos:
This is intended to repair #652, which I believe is an actual bug. The upper-bound check was missing on the unsigned version of the conversion routine (it was already present for the signed case).

### DIFF
--- a/include/boost/multiprecision/cpp_bin_float.hpp
+++ b/include/boost/multiprecision/cpp_bin_float.hpp
@@ -1631,13 +1631,9 @@ inline void convert_to_unsigned_int(I* res, const cpp_bin_float<Digits, DigitBas
 
    if (arg.sign())
    {
-      #ifdef BOOST_HAS_INT128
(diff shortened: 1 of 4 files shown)


jzmaddock:
Thanks Chris, yes your fix looks correct to me!

jzmaddock:
Ugh, always another corner case!  I guess since we saturate, then zero would be the logical choice?

If they aren't there already, we should have some consistent tests (by which I mean that are run for all the integer backends) for all this.  I can't remember off the top of my head where I put conversion tests (for saturation especially) - whether they are in test_arithmetic.hpp with everything else, or in the separate conversion testers.

ckormanyos:
> Ugh, always another corner case! I guess since we saturate, then zero would be the logical choice?

Indeed! OK, I'm working on it....

I just ran some tests on `cpp_dec_float`, `cpp_bin_float` and built-in `double`, such as [here at godbolt](https://godbolt.org/z/YE7PfWqrr).

I will make bin-float behave the same as dec-float and `double`. I'll go ahead and pack the code and its tests right in this PR while we are on such a roll.

PR #687, opened 2025-05-06
Adjust CircleCI setup to account for Boost.Inspect build change.

grafikrobot:
In addition to changing how to build the inspect tool this change Also simplifies the build and test to avoid future config problems as it can more easily adjust to future CI changes. It bumps the docker image to what the current stable image is to auto-adjust to recent stable gcc compiler.


jzmaddock:
Quickbook builds fine for me with gcc-13, so I'm not sure why it would be doing that.

jzmaddock:
This fix to Config gets it: https://github.com/boostorg/config/pull/517

Note that the point of the original comment, was that moving the dist-bin target to a different directory is a breaking change that breaks several libraries CI builds.

jzmaddock:
@grafikrobot : still seems to fail with the strange internal compiler errors :(

PR #691, opened 2025-05-27
Fix warning from scoped_default_precision

pps83:


pps83:
`precision_guard` is created in many places, it's too much mess to mark it unused for each var. Making struct [[maybe_unused]] seems to be working

PR #700, opened 2025-06-24
Update Github Actions and remove deprecated runners

mborland:


ckormanyos:
I think it ran fast on my branch in #701 in that PR. So maybe this 6 hours here was just a phantom

Issue #686, opened 2025-05-01
Array access warnings from cpp_bin_float


jzmaddock:
When building test_gamma_mp in Boost.Math with GCC-13 Mingw:

```
In file included from ..\..\../boost/multiprecision/cpp_int.hpp:2354,
                 from ..\..\../boost/multiprecision/cpp_bin_float.hpp:15,
                 from test_gamma_mp.cpp:19:
In function 'constexpr void boost::multiprecision::backends::right_shift_generic(Int&, boost::multiprecision::double_limb_type) [with Int = cpp_int_backend<151, 151, boost::multiprecision::unsigned_magnitude, boost::multiprecision::unchecked, void>]',
    inlined from 'constexpr typename std::enable_if<(! boost::multiprecision::backends::is_trivial_cpp_int<boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, boost::multiprecision::unsigned_magnitude, Checked, Allocator> >::value)>::type boost::multiprecision::backends::eval_right_shift(cpp_int_backend<MinBits, MaxBits, boost::multiprecision::unsigned_magnitude, Checked, Allocator>&, boost::multiprecision::double_limb_type) [with long long unsigned int MinBits1 = 151; long long unsigned int MaxBits1 = 151; boost::multiprecision::cpp_int_check_type Checked1 = boost::multiprecision::unchecked; Allocator1 = void]' at ..\..\../boost/multiprecision/cpp_int/bitwise.hpp:643:26,
    inlined from 'constexpr typename std::enable_if<(! boost::multiprecision::backends::is_trivial_cpp_int<boost::multiprecision::backends::cpp_int_backend<MinBits, MaxBits, boost::multiprecision::unsigned_magnitude, Checked, Allocator> >::value)>::type boost::multiprecision::backends::eval_right_shift(cpp_int_backend<MinBits, MaxBits, boost::multiprecision::unsigned_magnitude, Checked, Allocator>&, boost::multiprecision::double_limb_type) [with long long unsigned int MinBits1 = 151; long long unsigned int MaxBits1 = 151; boost::multiprecision::cpp_int_check_type Checked1 = boost::multiprecision::unchecked; Allocator1 = void]' at ..\..\../boost/multiprecision/cpp_int/bitwise.hpp:604:1,
[...]

PR #702, opened 2025-06-28
Suppress spurious array bound warning from gcc.

jzmaddock:
Fixes: https://github.com/boostorg/multiprecision/issues/686


ckormanyos:
Hi John, I've often found that suppressing array bounds also needs to be coupled with

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-overflow="
```

and then the required `pop`. I don't know if this will be the case here, but it woudln't surprise me either.

Cc: @mborland

jzmaddock:
Drone failure is a network issue, circleCI is quickbook failing to build for some strange reason... merging.

PR #707, opened 2025-06-30
Change CircleCI runner image to a "known good"

jzmaddock:


ckormanyos:
It's a mildly bigger discussion, but I think we could potentially start considering branch protection for Multiprecision. Maybe a bit down the road if this all stays stable.

Cc: @mborland

mborland:
Here it's not so much branch protection as it is protection from external variables. All the modular changes are merged into develop, but not yet into master. I expect to see some red that's not real red when we merge to master for next release.

ckormanyos:
Hi Matt, I meant more along the lines of the following, ...

- GHA has become quite fast and relatively reliable. This is based on a lot of work and quite honestly the depth of runner availability now through the C++ Alliance.
- Drone is running quickly and somewhat reliably.
- And CircleCI

So what I would like is the following. When I develop, I often times don't look at every single CI run. And I don't want to. I would like the meaning of green to actually mean GREEN.

So when I get green, I merge to develop. We are close to this quality level.

You won't like this, but I also would like GHA to run on push-to-branch _BEFORE_ I go for the PR. I'm often hitting real rookie errors in the first PR run. And I should have swept those out on a quick GHA run prior to really raising the PR.

Cc: @jzmaddock

mborland:
Couldn't rookie errors be determined by running the tests locally before push? I don't think we should be DDoSing shared resources like that. I run a few docker containers with various architectures that plug directly in my IDE (Clion). It's a trivial dropdown to change what I'm running on.

image

mborland:
> 
> To be direct, I never really figured out how to do that. And just (lazily) mis-used the runners. If you could walk me through it, ... then problem solved.
> 

Bring your laptop and some Dinkelacker Privat with you when you come visit, then I'm sure we can get it working.

PR #708, opened 2025-07-01
More cover dec_float and improve comments

ckormanyos:


ckormanyos:
OK so this was a productive, exciting and fun one.

- We got a couple $0.1\\%$ more coverage.
- Picked up a lot of edge-cases in `cpp_dec_float` and als `cpp_bin_float`.
- I found a place I can strengthen `cpp_double_fp_backend` down the road. But it's not a showstopper in any way.

So I'm goint to commit this. We didn't _quite_  get up to $95.0\\%$ coverage, but not bad at $94.7\\%$, which is up from $94.1\\%$ compared to 1.88.

Good progress.

So for me, it's time to ramp down for 1.89. I'm done (or done in).

Cc: @jzmaddock and @mborland

PR #709, opened 2025-07-01
Simplify and update Markdown

ckormanyos:


ckormanyos:
So this finishes my big effort on 1.89

PR #710, opened 2025-07-02
Merge to Master for 1.89

mborland:


ckormanyos:
Thank you very much Matt (@mborland).

Looking at the change log, this release seems like it really is a big release.

We made it. Yeah!

ckormanyos:
Thank you Matt, i agree with your assessment.

I must admit, right at the finish line Drone crumbles. This is non-reliable. I hope it can get better some day. As a developer, we rely on the meaning of green. Green should actually mean green and red should mean not green. Unreliable runners mean, ... I dunno.

It puts us in a difficult position when the runners have such unreliability.

Issue #703, opened 2025-06-28
Possible optimize cpp_fp_double_backend single-limb add/sub/mul/div


ckormanyos:
The new `cpp_double_fp_backend` type is composed of two limbs.

At the present time, add/sub/mul/div always do full 2-by-2 limb operations. This issue investigates odd double-limb, single-limb operations like the double-fp type being multiplied by a single fp-type having the sam3e width as its constituents.

PR #711, opened 2025-07-03
Obtain more code coverage

ckormanyos:
This PR obtains more code coverage. But it is for later in 1.90 (not for today's 1.89).


ckormanyos:
Ok so I worked on coverage, mostly with focus on `cpp_dec_float`. That file is pretty well covered now, but missing 8 interesting lines. I'll let this one sit for a while.

I would like to improve the line coverage of the tests and then we can discuss which headers need more dedicated edge tests.

Let's just leave this as a draft until the action on 1.90 wraps up.

ckormanyos:
In `cpp_dec_float` going to nearly $99\\%$ coverage was straightforward. Sure it needed a few hundred lines of dedicated test code.

I only found 1 bug, but harmless, in the area of negative infinity ($\-\infty$). There were several areas having redundant or unclear lines. I cleared up (or subjectively think I cleared up) a bunch of these sequences. Nothing really critical was found, and this shows how well this thing was tested to begin with. I'm really impressed with the test-suite.

ckormanyos:
Hi @jzmaddock and @mborland

Check it out, WE WILL NOT merge this until AFTER 1.89.

I have now marked this as _ready for review_. With this, I mean _AFTER_ 1.89 gets out the door. But the scope of change is non-trivial.

So I saw a window of opportunity (timewise and energy-wise) to go for coverage. I got a solid $96.1\\%$, which is good.

OK so there were some highlights.

- I found a small handful of bugs in `cpp_dec_float` and in some special functions edge-cases.
- I think that some infinity/NaN edge-cases in `default_ops`-pow were incorrect (also `erf`and `erfc`, maybe `tgamma`).
- In general I think these things are more right now than earlier, but there is work to be done.
- With this, I got the special-functions C99 thing to actually use signed-zero on `cpp_bin_float` to get lots of coverage lines. I think it is _right_-er than before, but not yet fully done. 
- I leaned heavily into `cpp_dec_float` and got under 10 lines missing in coverage. I hit this one, changed quite a bit actually. But my confidence is high here.
- I did some trivial changes to `complex_adaptor` and its tests to gain lots of lines.
- Along those lines, I got `cpp_dec_float` working with `complex_adaptor`, which was previously not the case since some `eval_whatever` functions were actually missing from this backend.
- I got the test cases `issue464` and `issue167` actually working and contributing to coverage. These were simply exception-ing out previously.
- I found redundant code in a few places like `pow`functions and streamlined them.

Well, that's about it.

ckormanyos:
There is a lot of material within the scope of this PR. It might take a while to sift through it all and negotiate on the best (or better) fixes, as I might not have subjectively always chosen the right path.

But in general, I think this PR is intact. And we can start taking a serious look at it.

ckormanyos:
I think it is time to review this PR.

Hi @jzmaddock and @mborland now that 1.89 is out, I'm satisfied with this PR for the moment. This PR is all about coverage and some syntax. Some bugs were found and corrected.

The highlights (repeated from above) include:

- I found a small handful of bugs in `cpp_dec_float` and in some special functions edge-cases.
- I think that some infinity/NaN edge-cases in `default_ops`-pow were incorrect (also `erf`and `erfc`, maybe `tgamma`).
- In general I think these things are more right now than earlier, but there is work to be done.
- With this, I got the special-functions C99 thing to actually use signed-zero on `cpp_bin_float` to get lots of coverage lines. I think it is _right_-er than before, but not yet fully done. 
- I leaned heavily into `cpp_dec_float` and got under 10 lines missing in coverage. I hit this one, changed quite a bit actually. But my confidence is high here.
- I did some trivial changes to `complex_adaptor` and its tests to gain lots of lines.
- Along those lines, I got `cpp_dec_float` working with `complex_adaptor`, which was previously not the case since some `eval_whatever` functions were actually missing from this backend.
- I got the test cases `issue464` and `issue167` actually working and contributing to coverage. These were simply exception-ing out previously.
- I found redundant code in a few places like `pow`functions and streamlined them.

ckormanyos:
OK this one is green again and ready for review.

I'm pretty satisfied with these results. Project coverage is at a solid $96.2\\%$, with `cpp_double_fp_backend.hpp` fully covered and `cpp_dec_float` missing fewer than $5$ lines.

A few bug fixes were done along the way. In future work, I'd like to see how to hit some of the missing lines in `cpp_int.hpp` and `cpp_bin_float.hpp`. I took a few tries at this but only had modest success.

When stressing `cpp_bin_float`, I ran across #726, which seems like an actual (still non-solved) bug.

There is also a hefty TODO list (for myself) at #703. So after concluding this `more_coverage` PR, I'll move on to #703 and possibly #726.

Cc: @jzmaddock and @mborland

include/boost/multiprecision/detail/default_ops.hpp
```diff
@@ -2616,14 +2619,29 @@ inline BOOST_MP_CXX14_CONSTEXPR typename multiprecision::detail::expression<tag,
 template <class Backend, multiprecision::expression_template_option ExpressionTemplates>
 inline BOOST_MP_CXX14_CONSTEXPR multiprecision::number<Backend, ExpressionTemplates> erf BOOST_PREVENT_MACRO_SUBSTITUTION(const multiprecision::number<Backend, ExpressionTemplates>& arg)
 {
+   const int fpc_arg { (boost::multiprecision::fpclassify)(arg) };
```
jzmaddock: Using fpclassify here is a bit heavyweight just to test for zero, using `arg.is_zero()` would be better, but actually, why do we need this at all - surely boost::math::erf should handle this already?

include/boost/multiprecision/detail/default_ops.hpp
```diff
@@ -2616,14 +2619,29 @@ inline BOOST_MP_CXX14_CONSTEXPR typename multiprecision::detail::expression<tag,
...
    return boost::math::erf(arg, c99_error_policy());
 }
 template <class tag, class A1, class A2, class A3, class A4>
 inline BOOST_MP_CXX14_CONSTEXPR typename multiprecision::detail::expression<tag, A1, A2, A3, A4>::result_type erf BOOST_PREVENT_MACRO_SUBSTITUTION(const multiprecision::detail::expression<tag, A1, A2, A3, A4>& arg)
 {
    using value_type = typename multiprecision::detail::expression<tag, A1, A2, A3, A4>::result_type;
-   detail::scoped_default_precision<value_type>                                          precision_guard(arg);
+   detail::scoped_default_precision<value_type> precision_guard(arg);
+
+   const int fpc_arg { (boost::multiprecision::fpclassify)(arg) };
```
jzmaddock: As above, but this is even worse, since we evaluate the same expression template *twice* here, once for fpclassify and once for the erf call.

include/boost/multiprecision/detail/default_ops.hpp
```diff
@@ -2675,12 +2693,27 @@ inline BOOST_MP_CXX14_CONSTEXPR typename multiprecision::detail::expression<tag,
...
       errno = ERANGE;
       return 1 / arg;
    }
+   else if ((fpc_arg == FP_NAN) && std::numeric_limits<local_number_type>::has_quiet_NaN)
+   {
+      return arg;
+   }
+   else if ((fpc_arg == FP_INFINITE) && std::numeric_limits<local_number_type>::has_infinity && std::numeric_limits<local_number_type>::has_quiet_NaN)
+   {
+      const bool is_neg { arg.compare(local_number_type(0)) < 0 };
```
jzmaddock: `arg.sign() < 0` would be more efficient.

As before I'm wondering why tgamma doesn't already handle this, and/or if there might not be a bug somewhere else?

include/boost/multiprecision/detail/functions/pow.hpp
```diff
@@ -499,27 +500,29 @@ inline void eval_pow(T& result, const T& x, const T& a)
...
+
    if ((a.compare(si_type(1)) == 0) || (x.compare(si_type(1)) == 0))
    {
       result = x;
       return;
    }
-   if (a.compare(si_type(0)) == 0)
+
+   const int fpc_x { eval_fpclassify(x) };
+   const int fpc_a { eval_fpclassify(a) };
```
jzmaddock: I don't think fpc_a is used anywhere else?  If not then comparison to zero should be faster, but even better would be `eval_is_zero`.

jzmaddock:
Thanks Chris for this, looks good other than my few comments on the header changes.

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: Understood. Thank you. I will repair this.

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: Understood. Thank you. I will repair this.

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: > As before I'm wondering why tgamma doesn't already handle this, and/or if there might not be a bug somewhere else?

I will look into this. I was fiddling around with `tgamma` because of coverage in the C99-function compatibility tests. Maybe a focussed on the wrong place. I will investigate this prior to merging the PR.

include/boost/multiprecision/detail/functions/pow.hpp
ckormanyos: Understood. Thank you. I will repair this.

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: Hi John (@jzmaddock) I finally had a moment and started looking into these edge cases.

In my branch, I have activated the `test_c99_appendix_F()` tests for both `cpp_double_double` as well as `cpp_bin_float` (which had previously not been the case). This all happens in the file `test_sf_import_c99.cpp` in Multiprecision's vast suite of tests.

The problem arises in the special function test that tests negative 0 in the error function [in my branch here](https://github.com/boostorg/multiprecision/blob/2f737ffa6372582b3f9c9ccea9b2ee7b8b8764ad/test/test_sf_import_c99.cpp#L1835).

When doing this, the [_generic_ implementation](https://github.com/boostorg/math/blob/43239024883d0f4986bdfce4cb471b3563faf161/include/boost/math/special_functions/erf.hpp#L147) of `erf_imp()` is called.

When `erf(cpp_bin_float(-0.0))` is called, the generic implementation does not catch the negative zero. It simply goes to the Taylor series and returns zero.

So I corrected a problem, I just corrected it in the wrong place. We actually need to add a very few lines to [`erf_imp()`](https://github.com/boostorg/math/blob/43239024883d0f4986bdfce4cb471b3563faf161/include/boost/math/special_functions/erf.hpp#L147).

ckormanyos: I am happy to take a try at the changes in Math's specfun.

jzmaddock:
>When erf(cpp_bin_float(-0.0)) is called, the generic implementation does not catch the negative zero. It simply goes to the Taylor series and returns zero.

Good catch!

My gut says something like:

```
if(fabs(z) < tools::root_epsilon<T>())
   return invert ? 1 : 2 * z / constants::root_pi<T>();
```

Which should make that zone more efficient as well as fixing the signed zero issue?  Note: just typed in off the top of my head!!

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: I'm working out a sensible repair in Math, partly based on your suggestion below. Thanks @jzmaddock

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: Resolved above

include/boost/multiprecision/detail/default_ops.hpp
ckormanyos: I simply reverted the `tgamma` changes. I do not (locally) see any more need from them. Perhaps I was initially mistaken. I'll see how CI/CD run in both Math and Multiprecision and repair anything if still present.

include/boost/multiprecision/detail/functions/pow.hpp
ckormanyos: Actually, the newly introduced `fpc_a` and `fpc_x` are used below, both of them a few times. Indeed, we could discuss if and when it makes sense to pull out the heavyweight `eval_fpclassify`.

For the exact change above, I reverted to `eval_is_zero(a)` for the common case of zero exponent which returns result 1 (even for NaN). That is a good, fast catch.

ckormanyos:
Removed the `constexpr`-ness efforts, as these will be handled in a future, separate branch. The most modern, mostly passing commit having the bulk of the `constexpr`-ness efforts is ac1b9545e7775bfb7918f5558f74567a897bbe93.

Issue #717, opened 2025-07-10
Collective Documentation Corrections


ckormanyos:
This issue collects proposed documentation corrections in the below listed issues.

- [x] #694
- [x] #712
- [x] #713
- [x] #714
- [x] #715
- [x] #716
- [x] #718
- [x] #723

I'm a bit busy this week, but might have a slot on the weekend for these.

Cc: @jzmaddock and @mborland and @LegalizeAdulthood

ckormanyos:
See also #720 in relation to #716

Issue #715, opened 2025-07-08
Markup formatted instead of interpreted


LegalizeAdulthood:
See https://www.boost.org/doc/libs/latest/libs/multiprecision/doc/html/boost_multiprecision/tut/interval.html

The bottom of the page has the markup `[ section:mpfi mpfi_float]` displayed instead of processed/interpreted into documentation.

ckormanyos:
Good catch Richard. I'm not entirely sure if we can make 1.89, but then the next one for sure.

jzmaddock:
Just a note that we can make documentation changes right up till release.

ckormanyos:
> Just a note that we can make documentation changes right up till release.

Thanks John for pointing that out. See also #717.

ckormanyos:
See also #719

Issue #712, opened 2025-07-08
documentation makes stale reference to GMP 5.1


LegalizeAdulthood:
> note also that the [mpf_class](http://gmplib.org/manual/C_002b_002b-Interface-Floats.html#C_002b_002b-Interface-Floats) wrapper that will be supplied with **GMP-5.1** reduces the number of temporaries to pretty much zero

https://www.boost.org/doc/libs/latest/libs/multiprecision/doc/html/boost_multiprecision/intro.html

GMP is currently at 6.3.0, so I think talking about 5.1 in the future sense is stale.

ckormanyos:
Good catch Richard. I'm not entirely sure if we can make 1.89, but then the next one for sure.

ckormanyos:
See also #719

PR #719, opened 2025-07-10
Fix a couple doc issues

LegalizeAdulthood:
I couldn't fix the one about the visualizer, because I don't know where it lives currently.

Relates to #717


ckormanyos:
> I couldn't fix the one about the visualizer, because I don't know where it lives currently.

I can't even find the sandbox. Maybe we should simply re-word this and consider it and describe it as a legacy topic. I do, however, think that this topic can be modernized. Maybe one day, really down the road, we can look into a visualizer.

There is a typo in the first sentence.

"Lets face it, debugger" ...

Should be 

"Let's face it debugging" ...

ckormanyos:
OK @LegalizeAdulthood I decided to finish this one to get #712 and #715 in there. I'll take a crack at the outdated page ASAP.

Issue #720, opened 2025-07-11
Modernize Multiprecision's Visualizers in VS


ckormanyos:
The purpose of this issue is to track attempted modernization of Multiprecision's Visualizers in VS. The visualizer development in MSVC has been modernized by the supplier. It seems to use managed `C#`. It should be feasible to start with `cpp_dec_float` and glue-code for `C#`, although the template parameters might be a bit tricky to work out. If this works out, we can try `cpp_bin_float`.

I can take a try at this one myself.

See also #716

Cc: @jzmaddock and @LegalizeAdulthood

PR #721, opened 2025-07-11
Modernize visualizer doc and maybe future plan

ckormanyos:


ckormanyos:
So I took a stab at modernizing the Visualizer docs.

Any comments on how I reworded the section in this PR?

For the future, which is not now, I even created #720 and believe I can chew through the necessary `C#` to modernize at least one visualizer. I've done a few things like this in the past. But uniting MP's radical templates with managed code may prove challenging.

jzmaddock:
IMO the old visualizers are basically obsolete, maybe just remove for now?

PR #736, opened 2025-09-07
Add more cpp_double_fp tests and handle warns

ckormanyos:


ckormanyos:
The purpose of this PR is to handle warnings in `cpp_double_fp_backend` and add a few more tests.

PR #746, opened 2025-10-25
Updated GCC on MinGW needs mbig-obj

ckormanyos:


mborland:
Once this is merged it'll get automatically pulled into #745 and the Master CI will re-run

ckormanyos:
The drone failure is a spurious glitch setting up the runner. Merging.

PR #756, opened 2026-04-07
Attempt constepr-ification of hash functions

ckormanyos:


ckormanyos:
Looks like this one can't work. It appears as though `std::hash` was never intended to be `constexpr`.

PR #758, opened 2026-04-15
Use fixed seed double-fp heats versus others

ckormanyos:


ckormanyos:
See also #757

ckormanyos:
Failure is spurious runner on drone. Merging

PR #760, opened 2026-04-30
Fix non-finite return value of abs(complex(float128))

### LINKED ISSUE #759: abs on an infinite complex float128 returns nan instead of inf
`abs` on an infinite `complex` `float128` returns `nan` instead of `inf`. 

Using Boost version 1.83 and g++ (GCC) 15.2.1 20260123 on Fedora Linux 42...

Here's `cmplxabs.cpp`
```c++
#include <iostream>
#include <complex>
#include <limits>

typedef boost::multiprecision::float128 REAL;

int main() {
  using std::abs;
  REAL x = std::numeric_limits<REAL>::infinity();
  std::cout << "abs(" << x << ") = " << abs(x) << "\n";
  std::complex<REAL> z = x;
  std::cout << "abs(" << z << ") = " << abs(z) << "\n";
}
```

Compiling and running with `g++ -o cmplxabs cmplxabs.cpp -lquadmath && ./cmplxabs` gives
```
abs(inf) = inf
abs((inf,0)) = -nan
```

mborland:
Closes: #759

### DIFF
--- a/include/boost/multiprecision/float128.hpp
+++ b/include/boost/multiprecision/float128.hpp
@@ -20,6 +20,7 @@
 #include <cfloat>
 #include <tuple>
 #include <cstring>
+#include <complex>
 #include <boost/multiprecision/detail/standalone_config.hpp>
 #include <boost/multiprecision/number.hpp>
 #include <boost/multiprecision/detail/hash.hpp>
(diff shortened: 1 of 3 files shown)


cffk:
Thanks!

